PR #20, opened 2020-03-03
Add NotifyServerLevelProcessingComplete GameMessage

elenawinters:
Add NotifyServerLevelProcessingComplete GameMessage

I'll refine it later. There is an issue when using the /level command, where it'll kind of spam chat. I'm working to fix this.


Jettford:
This GM does nothing as far as I am aware

elenawinters:
@Jettford This is the GM that the client sends when the character levels up.

Jettford:
Still unused though

PR #35, opened 2020-07-26
Added link to win10 setup blog for redis in README

Bitl:
I'll probably improve on this in the morning. Did it on phone in the middle of the night so it could be messed up a bit.


Jettford:
I have already suggested the following change @yuwui, they thought that was it obvious and not worth changing.

Bitl:
I guess I'll close this.

PR #82, opened 2020-10-07
Skill syncing

MickVermeulen:
Refactor of the entire skill system. Skill trees are executed in two steps: first a (de-)serialization step and then and execute step. The (de-)serialization handles the reading or writing of the behavior GMs and prepares the tree to be executed using a set of execution parameters, EchoStartSkill and EchoSyncSkill are then sent immediately after (de-)serialization to allow very fast communication between client and server. Once the tree is ready to be executed it can be executed as a background process.

### DIFF
--- a/Uchu.World/Enums/MovementType.cs
+++ b/Uchu.World/Enums/MovementType.cs
@@ -2,11 +2,13 @@ namespace Uchu.World
 {
     public enum MovementType : uint
     {
-        Ground = 1,
+        Moving,
+        Ground,
         Jump,
         Falling,
         DoubleJump,
-        Jetpack = 6,
-        Stunned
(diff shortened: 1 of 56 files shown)


Wincent01:
Does not have anything todo with the `parameters.Context.ExplicitTarget`, if `UsePickedTarget` is true and a target exists, it branches from there. The validity of a target is checked on sync.
https://github.com/yuwui/Uchu/blob/skill-syncing/Uchu.World/Systems/Behaviors/TacArcBehavior.cs#L45
https://github.com/yuwui/Uchu/blob/skill-syncing/Uchu.World/Systems/Behaviors/TacArcBehavior.cs#L166

PR #89, opened 2020-10-12
Damage calculation

MickVermeulen:
Fixes multiple issues:

- Damage calculation is redone
- Race conditions in the packet handling and sync skill handling
- AOE attacks


Jettford (approved):
Can’t see any issues and it is nice you realised the packet handler issue when you did

PR #96, opened 2020-10-17
Attempt at fixing 55% bug

Jettford:


MickVermeulen:
Don't think this is the fix bc this method instructs the new player about the game objects in the world, not the other way around. Also, destroying the player here might cause unexpected behavior in other parts of Uchu as they might rely on the old player still being around. The fix should be properly handling the destruction of the player when leaving the world, not handling that bug afterwards using code.

PR #103, opened 2020-10-18
Block client commands being responded to by the server

Jettford:


MickVermeulen:
So two small things: I know I didn't tell you this before but I think it might be better to have the client commands array as a static property on the SocialHandler, that way it's not built every time this is called, so `private static string[] ClientCommands { get; } = {""};`, also it might be nice to format the client commands using newlines a bit so that they don't cross the 120 character mark, makes it easier to read which commands there are without scrolling.

Jettford:
Oh yeah you did tell me, I will do that now

Jettford:
Fixed, going to merge

Issue #91, opened 2020-10-15
drop client root seems to be sent slowly


MickVermeulen:
It seems like after a DieMessage the DropClientLoot message is sent really slowly. This causes issues because it can provide the idea of lag while there actually isn't a lot of lag, the issue is just that the server doesn't send the message quick enough.

MickVermeulen:
Probably related to game loop optimizations and reducing concurrency, being worked on in another branch.

MickVermeulen:
Fixed in #110

PR #110, opened 2020-11-01
Gameloop performance

### LINKED ISSUE #109: Game loop performance
The game loop is currently quite slow due to misconfiguration of certain tasks.

MickVermeulen:
Closes #109 and #91 . Introduces new issue where behavior sync serialization is messed up and thus projectiles won't show up but due to the scope of this PR I will fix that in a subsequent issue. In this subsequent issue I'll also remove the Reader and Writer from the ExecutionContext and add them as arguments to behaviors as having these as shared parameters has proven to easily create race conditions.


Jettford (approved):
This looks good and boosts the performance significantly but I would like @Wincent01 to also review as he may spot something I can't.

PR #136, opened 2020-11-20
Fix Typo Error Message for a null level

TheMatt2:
Make it so a null level returns a message saying "level null" not "username null".

This can be tested by pulling from the API 

    curl "localhost:10000/account/level?u=name&l="


MickVermeulen:
Looks good, nice that you added the newline as well. Merging now.

PR #137, opened 2020-11-22
Command System refactor

Jettford:


MickVermeulen:
Like discussed, don't remove the Enum but set the enum values to the ones that were used in-game.

Jettford:
Yeah

PR #140, opened 2020-11-25
Enhancement/smash lag spikes

### LINKED ISSUE #119: smashing leads to cpu spikes
Smashings destructibles causes huge lag spikes, mission checks are also very slow in the background.

MickVermeulen:
closes #119 

Quite a large PR that refactors all possible components of the mission system that might be cached. Missions are now rarely fetched from the database, only when progress is updated. Mission items are now also re-enabled due to it no longer being a performance issue.


Jettford (approved):
After testing all the changes they improve the speed as intended

PR #144, opened 2020-11-29
GameMasterLevel fixed along with commands renamed

Jettford:


Uchu.World/Handlers/WorldInitializationHandler.cs
```diff
@@ -212,13 +217,14 @@ private async Task SendCharacterXmlDataToClient(Character character, IRakConnect
         /// The generated XML data is based on https://docs.google.com/document/d/1XDh_HcXMjSdaGeniG1dND5CA7jOFXIPA_fxCnjvjaO4/edit#
         /// </remarks>
         /// <param name="character">The character to generate the XML data for</param>
+        /// <param name="GMLevel">The GM level of the user</param>
         /// <returns>XmlData conform with the LU Char Data XML Format</returns>
-        private static XmlData GenerateCharacterXmlData(Character character)
+        private static XmlData GenerateCharacterXmlData(Character character, int GMLevel)
```
MickVermeulen: Use lowercase parameter names (e.g. gmLevel)

MickVermeulen (commented):
I like the fact that you added separate roles for the mythran and mythran with glow, just had a minor comment about the style you used for a parameter. Probably would be nice to have aliases for the admin commands as I think that testmap and gmadditem are really weird names (although being the names used in live).

Jettford:
> Probably would be nice to have aliases for the admin commands as I think that testmap and gmadditem are really weird names (although being the names used in live).

testmap and gmadditem are the standard, the world and give commands are the odd ones out.

Uchu.Core/Enums/GameMasterLevel.cs
```diff
@@ -2,10 +2,11 @@ namespace Uchu.Core
 {
     public enum GameMasterLevel
     {
-        Player,
-        Mythran,
-        Admin,
-        Operator,
+        Player = 0,
+        Mythran = 1,
+        MytranWithGlow = 2,
```
enteryournamehere: Typo in MytranWithGlow, should be MythranWithGlow

Uchu.World/Handlers/WorldInitializationHandler.cs
Jettford: Done

MickVermeulen (approved):
Looks good 👍

Uchu.World/Handlers/WorldInitializationHandler.cs
```diff
@@ -191,13 +194,15 @@ private async Task SendCharacterXmlDataToClient(Character character, IRakConnect
             var xml = new byte[bytes.Length - 3];
 
             Buffer.BlockCopy(bytes, 3, xml, 0, bytes.Length - 3);
-
+            
+            var template = gmlevel >= 2 && gmlevel < 5 ? 10103 : 1;
```
enteryournamehere: seeing the enum doesn't include a level with 3 or 4 as value, why not write this as
`var template = gmlevel == GameMasterLevel.MythranWithGlow ? 10103 : 1`?

Jettford: oh yeah

Uchu.Core/Enums/GameMasterLevel.cs
Jettford: Fixed

Uchu.World/Handlers/WorldInitializationHandler.cs
MickVermeulen: Good catch, I like this solution :)

Uchu.World/Handlers/WorldInitializationHandler.cs
enteryournamehere: @Jettford the `param` in the documentation/comment still has GMLevel instead of gmLevel (line 220 and 301)

Uchu.World/Handlers/WorldInitializationHandler.cs
enteryournamehere: Missing a semicolon after this line now. And could you rename `gmlevel` to `gmLevel` for consistency?

MickVermeulen:
Looks good, once @enteryournamehere approves the merge feel free to merge 👍

Uchu.World/Handlers/WorldInitializationHandler.cs
```diff
@@ -180,8 +180,11 @@ public async Task PlayerLoadedHandler(PlayerLoadedMessage message, Player player
         /// <param name="session">The session cache for the connection</param>
         private async Task SendCharacterXmlDataToClient(Character character, IRakConnection connection, Session session)
         {
+            await using var ctx = new UchuContext();
+            var gmLevel = ctx.Users.Where(i => i.Id == character.UserId).First().GameMasterLevel;
+            
             // Get the XML data for this character for the initial character packet
-            var xmlData = GenerateCharacterXmlData(character);
+            var xmlData = GenerateCharacterXmlData(character, gmlevel);
```
MickVermeulen: This will crash now bc the variable name is incorrect.

Jettford: oh for

Uchu.Core/Handlers/Commands/StandardCommandHandler.cs
```diff
@@ -290,13 +290,20 @@ public static async Task<string> SetGameMasterLevel(string[] arguments)
             {
                 return "Invalid <level>";
             }
+            
+            if (Enum.IsDefined(typeof(GameMasterLevel), int.Parse(arguments[1])))
```
enteryournamehere, line 294: Seems redundant to have both this check at line 289:
```cs
if (!Enum.TryParse<GameMasterLevel>(arguments[1], out var level))
{
    return "Invalid <level>";
}
```
and this one:
```cs
if (Enum.IsDefined(typeof(GameMasterLevel), int.Parse(arguments[1])))
{
    // ...
}
else
{
    return "Invalid <level>";
}
```

enteryournamehere (dismissed):

Uchu.Core/Handlers/Commands/StandardCommandHandler.cs
Jettford: One checks if the number is a int and the other checks if it is in the enum

enteryournamehere: My bad, it's not redundant, I misunderstood how TryParse works. But [TryParse](https://docs.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=net-5.0#code-try-6) will also work with the name of the enum value like "Mythran", which will give `level` the correct value, but then fail in the next conditional as the argument "Mythran" can't/won't be parsed as int. So I think that second if statement should be 
```cs
if (Enum.IsDefined(typeof(GameMasterLevel), level))
{
    // ...
}
```

TheMatt2: This does seem a bit redundant.
If a number is specified that is outside the valid range [TryParse](https://docs.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=net-5.0) will _succeed_, and convert `arguments[1]` to `level`.
So the `int.Parse(arguments[1])` seems redundant, as `level == int.Parse(arguments[1])` if `arguments[1]` is a number.

Also, if the world "Console" is specified, `Int.Parse` would error, which would be an issue.

The two exception cases can both be combined for simplicity.

So I propose this check should be used instead

```c#
if (!Enum.TryParse<GameMasterLevel>(arguments[1], out var level) || 
    !Enum.IsDefined(typeof(GameMasterLevel), level))
{
    return "Invalid <level>";
}

user.GameMasterLevel = (int) level;

...
```

TheMatt2: An similar case needs to be solve here: https://github.com/UchuServer/Uchu/blob/3421c7079bc4927ec8a538e88b3526c12d30c3ee/Uchu.Master/Api/AccountCommands.cs#L133

To prevent invalid enum values set via the API.

TheMatt2 (changes requested):
Potential Issue that would cause error if an Enum is specified as text, not number.

MickVermeulen:
I get @TheMatt2 's point and I agree that it's an even neater solution but we've been discussing this rather simple change for the entire day now, I suggest merging this and then @TheMatt2 submitting a subsequent PR regarding the change he thinks works better.

Jettford:
> You agreeded that it wasn't redundant

I can't english

enteryournamehere:
Just missing this https://github.com/UchuServer/Uchu/pull/144#discussion_r532190310

MickVermeulen:
Yeah @Jettford just update the docs and merge then, @TheMatt2 you can then submit another PR regarding your improved proposition so we can get this in dev now.

PR #145, opened 2020-11-29
Further Correct GameMasterLevel Handling

TheMatt2:
Corrects a new error that is caused by trying to specify the text of an enum at the console. This corrects it by properly parsing the text.

Also added handling to check GameMasterLevel values coming from the API


Jettford:
You win

PR #146, opened 2020-11-29
Make scripts zonespecific

### LINKED ISSUE #115: world scripts don't use ZoneSpecific
There's a few native scripts that don't use the ZoneSpecific annotation, as a result these scripts are loaded for all worlds. For example when loading up VE many AG scripts are also loaded.

enteryournamehere:
Resolves #115 

All scripts in the individual world folders (namespaces) now have their zone specified, except the TurretMech and SalutingNpc ones as they would apply to objects spawned in other worlds.


Jettford:
This is good to merge

PR #151, opened 2020-12-23
Fix major issues and perms

elenawinters:
- Fixed Character component (you can see other people's equipment and also equip things again)
- Add a /execute command for admins (this still has issues, consider it experimental)
- Coins with 0 value (showing up as a question mark) no longer drop
- Fixes the Teleport GameMessage, as well as the command.
- Fix permissions on /smash
- Fix permissions on /pvp
- Add .bat files to gitignore (I use bat files to build and run the server)

### DIFF
--- a/Uchu.World/Handlers/Commands/CharacterCommandHandler.cs
+++ b/Uchu.World/Handlers/Commands/CharacterCommandHandler.cs
@@ -8,6 +8,7 @@ using System.Numerics;
 using System.Runtime.CompilerServices;
 using System.Text;
 using System.Threading.Tasks;
+using System.IO;
 using Uchu.Core;
 using Uchu.Core.Client;
 using Uchu.Core.Resources;
@@ -86,6 +87,7 @@ namespace Uchu.World.Handlers.Commands
         {
             if (arguments.Length != 1) return "coin <delta>";
 
+            /// We parse this as an int instead of long, due to the max long causing bugs. Currency maxes out at a long.
             if (!int.TryParse(arguments[0], out var delta) || delta == default) return "Invalid <delta>";
 
             player.Currency += delta;
@@ -170,7 +172,7 @@ namespace Uchu.World.Handlers.Commands
             return $"{player.Transform.Rotation}";
         }
 
-        [CommandHandler(Signature = "smash", Help = "Smash yourself", GameMasterLevel = GameMasterLevel.Admin)]
+        [CommandHandler(Signature = "smash", Help = "Smash yourself", GameMasterLevel = GameMasterLevel.Player)]
         public async Task<string> Smash(string[] arguments, Player player)
         {
             var animation = "violent";
@@ -759,7 +761,7 @@ namespace Uchu.World.Handlers.Commands
 
             return $"Successfully added {type.Name} to {player}";
         }
-        
+
         [CommandHandler(Signature = "testmap", Help = "Transfer to world", GameMasterLevel = GameMasterLevel.Mythran)]
         public string World(string[] arguments, Player player)
         {
@@ -777,9 +779,9 @@ namespace Uchu.World.Handlers.Commands
                 return $"Can't find world with ID {id}";
             }
 
-            string WorldName = WorldTable.ZoneName;
+            var path = Path.Combine(UchuServer.Config.ResourcesConfiguration?.GameResourceFolder, Path.Combine("maps", WorldTable.ZoneName.ToLower()));
 
-            if (WorldName.EndsWith(".luz"))
+            if (File.Exists(path) && WorldTable.LocStatus > 0)
             {
                 //
                 // We don't want to lock up the server on a world server request, as it may take time.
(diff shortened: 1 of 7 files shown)


Uchu.World/Handlers/Commands/CharacterCommandHandler.cs
```diff
@@ -86,6 +87,7 @@ public string ChangeCoin(string[] arguments, Player player)
         {
             if (arguments.Length != 1) return "coin <delta>";
 
+            /// We parse this as an int instead of long, due to the max long causing bugs. Currency maxes out at a long.
```
MickVermeulen: Really nitpicky but we generally only use double slash comments inline and use triple slash comments for doc strings

Uchu.World/Handlers/Commands/CharacterCommandHandler.cs
```diff
@@ -777,9 +779,9 @@ public string World(string[] arguments, Player player)
                 return $"Can't find world with ID {id}";
             }
 
-            string WorldName = WorldTable.ZoneName;
+            var path = Path.Combine(UchuServer.Config.ResourcesConfiguration?.GameResourceFolder, Path.Combine("maps", WorldTable.ZoneName.ToLower()));
```
MickVermeulen, line 760: Just wondering, why was this buggy before and how does this fix it? Not that this is wrong or something, I'm just out of the loop.

Uchu.World/Handlers/Commands/OperatorCommandHandler.cs
```diff
@@ -236,5 +237,57 @@ public string Target(string[] arguments, Player player)
 
             return $"Target: {baseCombatAiComponent.Target}";
         }
+
+
+        [CommandHandler(Signature = "execute", Help = "Execute a command as another player", GameMasterLevel = GameMasterLevel.Admin)]
+        public async Task<string> Execute(string[] arguments, Player executor)
+        {
+            List<Player> players = new List<Player>();
+            if (arguments[0] == "*") {
```
MickVermeulen: The code in this block doesn't adhere to the overall Uchu / C# style, we generally do:

```
if ()
{
    ... stuff
}
else
{
    ... other stuff
}
```

and

```
foreach ()
{
    ... stuff
}
```

Uchu.World/Handlers/Commands/OperatorCommandHandler.cs
```diff
@@ -236,5 +237,57 @@ public string Target(string[] arguments, Player player)
...
+                if (player == default) return $"No player named {arguments[0]}";
+                players.Add(player);
+            }
+
+            await using var ctx = new UchuContext();
+            var character = await ctx.Characters.Include(c => c.User).FirstAsync(
+                c => c.Id == executor.Id
+            );
+
+            var _com = new List<string>(arguments);
```
MickVermeulen: Bit of a odd variable name, if you meant command I would just use `command` here

Uchu.World/Handlers/Commands/OperatorCommandHandler.cs
```diff
@@ -236,5 +237,57 @@ public string Target(string[] arguments, Player player)
...
+
+            var message = string.Join(" ", _com);
+
+            if (!message.StartsWith('/')) {
+                message = "/" + message;
+            }
+
+            if (!SocialHandler.ClientCommands.Contains(message.Split(" ").ElementAt(0)) && message.Split(" ").ElementAt(0) != "/execute" && message.Length > 1) {
+                foreach (Player player in players) {
+                    player.SendChatMessage("You feel a magical power...", PlayerChatChannel.Normal);
```
MickVermeulen, line 288: Lol I like this little extra

Uchu.World/Handlers/Commands/OperatorCommandHandler.cs
```diff
@@ -236,5 +237,57 @@ public string Target(string[] arguments, Player player)
 
             return $"Target: {baseCombatAiComponent.Target}";
         }
+
+
+        [CommandHandler(Signature = "execute", Help = "Execute a command as another player", GameMasterLevel = GameMasterLevel.Admin)]
+        public async Task<string> Execute(string[] arguments, Player executor)
```
MickVermeulen: I'd appreciate a doc string here with maybe some example usage

MickVermeulen (changes requested):
Looks good overall, just had some style / documentation requests.

Uchu.World/Handlers/Commands/CharacterCommandHandler.cs
Jettford: Basically the Zone tables are massive messes there are almost no ways to filter whether a world is correct or not but this does filter out most invalid zones

MickVermeulen: Seems fair.

MickVermeulen:
All looks good, if you tested the TP GM feel free to merge :)

MickVermeulen (approved):

PR #154, opened 2021-01-08
Add Authorization Port Configuration

### LINKED ISSUE #152: Make Master Server Port Configurable
Currently, the master server is hardcoded to use port 21836, which was due to the port being hardcoded in the RakNet DLL mod. The next release of the mod (https://github.com/lcdr/raknet_shim_dll/pull/1) will now allow for custom ports to be specified and use 21836 by default. Being able to change the master server port for Uchu, with the default of 21836, which would mean no specific ports would be required for a given deployment.

TheNexusAvenger:
Related (already closed) issue: https://github.com/UchuServer/Uchu/issues/152

This change allows the server owner to configure the port used by the Authorization server from the hardcoded 21836 to an arbitrary port. This may not get much use given it requires users to explicitly define the port and use the next release of lcdr's RakNet shim, but does reduce a hardcoded number that requires a code change if someone really needed a custom port.

The change was tested with the arbitrary port 20000 using the latest code from lcdr's RakNet shim repository using `cargo build --release` to build. It was also tested in the stock configuration (port 21836). Both cases allowed me to log in, select a character, and play in Avant Gardens.

### DIFF
--- a/Uchu.Core/Config/UchuConfiguration.cs
+++ b/Uchu.Core/Config/UchuConfiguration.cs
@@ -159,6 +159,11 @@ namespace Uchu.Core.Config
         [XmlElement] public string Hostname { get; set; } = "";
         
         /// <summary>
+        /// The port to run the authentication server at
+        /// </summary>
+        [XmlElement] public int AuthenticationPort { get; set; } = 21836;
+        
+        /// <summary>
         /// The port to run the character server at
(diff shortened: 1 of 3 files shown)


elenawinters (approved):

Jettford (approved):
Looks good

Issue #125, opened 2020-11-07
using  Guns doesnt drop loot


cool56571:

Jettford:
Is this with all ranged weapons or specific ones?

cool56571:
so far ive only tried using it with my flare guns but judging what ive been told its all projectiles

TheNexusAvenger:
I experimented with this a bit. The bug actually is that enemies don't take damage from guns. For some, this is due to a `NullReferenceException` when executing the behavior due to no behavior arguments being populated. I was able to resolve this on my build by adding `castType: SkillCastType.OnUse` in to line 42 in [`Projectile.ImpactAsync`](https://github.com/UchuServer/Uchu/blob/c16eba45265dc84e37cec189182998540b50e7f2/Uchu.World/Objects/Projectile.cs#L42) since the default `OnEquip` wouldn't populate the arguments. However, I have no clue if this is the actual fix because I don't know enough of the system.

Note: I confirmed this with the "Basic Flintlock Pistol" (6937), but the "Elite Long Barrel Blaster" (12729) has a different message (`Invalid behavior sync id: 7!`). The fix addresses some but not all.

Edit: For the `Elite Long Barrel Blaster`, it registers sync ids 2 and 6 but calls 3 for no Imagination and 7 for firing.

TheNexusAvenger:
This should be fixed with the latest changes to behaviors (pull request https://github.com/UchuServer/Uchu/pull/185), although it hasn't been tested with every item.

PR #156, opened 2021-01-11
Fix Turret Kills Not Dropping Loot

TheNexusAvenger:
There is no associated GitHub issue and does **_not close_** https://github.com/UchuServer/Uchu/issues/125.

I noticed while playing that when a turret kills a Stromling, the Stromling does not drop any loot. When debugging, the `parameters.Context.Associate` being read in [`BasicAttackBehavior`](https://github.com/UchuServer/Uchu/blob/c16eba45265dc84e37cec189182998540b50e7f2/Uchu.World/Systems/Behaviors/BasicAttackBehavior.cs#L99) was set to the Stromling instead of the Sentinel Turret (the buildable turret after killing a Stromling Mech). This is due to the `NpcExecutionContext` in the behavior tree serializing the `target` as the `associate`, which are not the same when the `target` is defined, like in `Projectile`. Based on the [constructor](https://github.com/UchuServer/Uchu/blob/c16eba45265dc84e37cec189182998540b50e7f2/Uchu.World/Systems/Behaviors/NpcExecutionContext.cs#L39), `associate` is what should be passed in, not `target`. This change corrects the input to the constructor to match the original design.

*I was instructed privately to include Wincent and Mick due to domain knowledge, as well as Jett for consulting on the change.*

### DIFF
--- a/Uchu.World/Systems/Behaviors/BehaviorTree.cs
+++ b/Uchu.World/Systems/Behaviors/BehaviorTree.cs
@@ -320,7 +320,7 @@ namespace Uchu.World.Systems.Behaviors
             CastType = SkillCastType.Default;
             target ??= associate;
 
-            var context = new NpcExecutionContext(target, skillId, syncId, calculatingPosition);
+            var context = new NpcExecutionContext(associate, skillId, syncId, calculatingPosition);
             if (!SkillRoots.TryGetValue(skillId, out var root))
             {
                 Logger.Debug($"Failed to find skill: {skillId}");


Jettford (approved):
After review of game play, this appear to improve the behaviors cast by the turrets

MickVermeulen:
This seems to remove the target from the behavior serialization, have you ensured that regular NPC attacks still function? For example stromlings attacking players and Epsilon attacking stromlings.

TheNexusAvenger:
> This seems to remove the target from the behavior serialization
Target is still used in the serialization but at another line.
https://github.com/UchuServer/Uchu/blob/1caccc83dd35da94b2d9f5f87b67ecd8a2ac14d3/Uchu.World/Systems/Behaviors/BehaviorTree.cs#L331

> have you ensured that regular NPC attacks still function? For example stromlings attacking players and Epsilon attacking stromlings.
Just tested again to make sure. Epsilon properly attacked several Stromlings and I was able to be smashed multiple times from both Stromlings and Mechs.

MickVermeulen:
Ah yes, looking at the overall code you implemented it properly. Quick explanation why this actually works: the behavior context (which now contains the associate) is the overall context for the entire behavior tree and should indeed contain the caster of the behavior. The behavior target, send in the parameters, is different for each branch in the tree and is thus still properly thrown around, nice catch!

MickVermeulen (approved):

PR #158, opened 2021-01-12
Improve Maelstrom Vacuum

### LINKED ISSUE #66: 55% load issue on AG
When a character leaves AG their character is not deconstructed, making it impossible for the player to rejoin. It seems to happen in AG specifically so it might be a script.

### LINKED ISSUE #80: paradox vacuum doesn't suck bricks
AG mission for Wisp Lee

TheNexusAvenger:
*Closes #66 and #80*

This pull request resolves a bug that the Maelstrom vacuum could only pick up 1 sample without rebooting, with the server tick processing sometimes stopping, as well as adding the missing animations for not being in range and actually collecting.

### DIFF
--- a/Uchu.StandardScripts/AvantGardens/MaelstromVacuum.cs
+++ b/Uchu.StandardScripts/AvantGardens/MaelstromVacuum.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Linq;
 using System.Numerics;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


Uchu.StandardScripts/AvantGardens/MaelstromVacuum.cs
```diff
@@ -11,30 +12,61 @@ public class MaelstromVacuum : NativeScript
...
             {
                 if (!(obj is AuthoredGameObject gameObject)) return;
                 
                 if (gameObject.Lot != 14596) return;
+                
+                if (!(gameObject.Author is Player player)) return;
 
-                var samples = Zone.GameObjects.Where(g => g.Lot == 14718).ToList();
+                Delegate onFireServerEventDelegate = null;
+                onFireServerEventDelegate = Listen(player.OnFireServerEvent, (name, message) =>
```
MickVermeulen, line 24: For these lines it's more idiomatic to use `var onFireServerEventDelegate = Listen( ... );`

Uchu.StandardScripts/AvantGardens/MaelstromVacuum.cs
```diff
@@ -11,30 +12,61 @@ public class MaelstromVacuum : NativeScript
...
+                        await Task.Delay(4000);
 
-                var smashable = selected.GetComponent<DestructibleComponent>();
+                        await smashable.SmashAsync(gameObject, gameObject.Author as Player);
+                    });
+                });
 
-                await smashable.SmashAsync(gameObject, gameObject.Author as Player);
+                Delegate onReadyForUpdatesEventDelegate = null;
+                onReadyForUpdatesEventDelegate = Listen(player.OnReadyForUpdatesEvent, (message) =>
```
MickVermeulen, line 60: The same as my previous comment applies here.

MickVermeulen (changes requested):
Just some minor style stuff. For the record: could you report a bit on what actually caused the issue? I presume it was serialization given that an animation was played that causes the server to be stuck at 55%?

Uchu.StandardScripts/AvantGardens/MaelstromVacuum.cs
TheNexusAvenger: The IDE I use suggests doing this, however, it needs to be referenced inside the lambda function as part of the current design so that the event listeners are cleared. Changing it to the recommended method results in a compile error because the lambda function does not see the delegate as a variable because the lambda function is used to create the delegate.

TheNexusAvenger:
> For the record: could you report a bit on what actually caused the issue? I presume it was serialization given that an animation was played that causes the server to be stuck at 55%?

¯\_(ツ)_/¯
I was informed it was the problem and was able to reproduce the problem went away with the fix in the first commit. I never dug into why it became a problem, and why in some cases the server tick would stop processing.

PR #173, opened 2021-01-16
Add Restarting Missions

### LINKED ISSUE #169: Missions Of The Day Don't Reset The Next Day
After completing the mission of the days and waiting for them to reset the next day, no repeatable missions become available the next day. Confirmed this through Epsilon, Outpost Console, and Rusty Steele.
image

TheNexusAvenger:
Closes https://github.com/UchuServer/Uchu/issues/169

This pull request adds the ability to restart completed missions, such as daily missions or helper missions in Nexus Tower. This also includes a fix for a bug I introduced with returning items in the Laser Rampage/Wrecking Crew not working due to a wrong condition operator.

The change was developed on `enhancement/reduce-load-time` and the changes were moved over to a different branch. The changes were tested by manually changing the completion time of a daily mission (id 1858 especially) in the database to be an older time to allow for the mission to be re-obtained. The mission was re-obtained, re-completed, and the incrementer in the database of how many times it completed went up.

### DIFF
--- a/Uchu.Core/Database/UchuContext.cs
+++ b/Uchu.Core/Database/UchuContext.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.ChangeTracking;
 using Uchu.Core.Providers;
 
 namespace Uchu.Core
@@ -58,6 +59,11 @@ namespace Uchu.Core
                     Logger.Error($"{config.Database.Provider} is a invalid or unsupported database provider");
                     throw new Exception($"Invalid database provider: \"{config.Database.Provider}\"");
             }
+        }
+        
+        public EntityEntry Entry(object entry)
+        {
+            return ContextBase.Entry(entry);
         }
         
         public async Task EnsureUpdatedAsync()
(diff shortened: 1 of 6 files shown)


Jettford (changes requested):
Changes to the CDClient are not a good idea

TheNexusAvenger:
The changes to CDClient were reverted. The intent with the change was to ensure that the cooldowns for daily missions would be valid while keeping the 4 repeatable missions in Nexus Tower at no cooldown as they are helper missions. However, I realized that those 4 missions are the *only* missions that are *not* in `InMOTD` with no cooldown. The case that corrects the time to 22 hours does not affect those missions, while fixing the missions in Avant Gardens.

MickVermeulen:
Nice addition, only downside would be your use of the UchuContext as that isn't required after the refractor is merged. Either way, you're done earlier than me so I'll port this to the refactor once I find the time.

TheNexusAvenger:
After some discussion in the LEGO Universe Community Hub, it was found that repeatable missions must use a mission state that is 8 + the normal mission state, which the current code does not do. There are a few missions preconditions that specify the 8 + (normal state) instead of just the normal state. This shouldn't be too bad, mainly changing `RestartAsync` to set a different state, but the change may be more involved.

MickVermeulen:
Thanks for the comment! I'll take that into account when I reimplement this.

MickVermeulen (approved):
Really nice work, no comments!

PR #174, opened 2021-01-17
Improve World Load Time

### LINKED ISSUE #117: World Loading is painfully slow

TheNexusAvenger:
*May or may not close https://github.com/UchuServer/Uchu/issues/117*

This pull request improves the load time of worlds. On my system, loading Avant Gardens went from roughly a minute to <15 seconds from world startup to the player loading in. This is done by adding an option to disable the caching service, which takes 10 seconds to time out if it isn't set up, as well as using client database caching that reduces the load times by roughly 35 seconds for Avant Gardens.

This change modifies a lot of code, including changing methods from asynchronous to synchronous. This may also alter the order of how things function and may result in unexpected changes to behavior. I haven't noticed anything after playing around for a few hours, but I may have missed something.


Uchu.Core/Logging/LogQueue.cs
```diff
@@ -0,0 +1,135 @@
...
+            return entry;
+        }
+
+        /// <summary>
+        /// Starts a thread for outputting the log. This prevents
+        /// the console IO from blocking the application.
+        /// </summary>
+        public void StartConsoleOutput()
+        {
+            new Thread(() =>
```
MickVermeulen: Why wouldn't you just Task.Run this? Then the ThreadPoolScheduler will determine the thread that you get, but it'll still run in the background.

Uchu.Core/Logging/LogQueue.cs
```diff
@@ -0,0 +1,135 @@
...
+                        // Write the entry.
+                        Console.WriteLine(entry.message);
+                        entry = this.PopEntry();
+                    }
+                    
+                    // Reset the color.
+                    Console.ResetColor();
+                    
+                    // Sleep for a bit before resuming.
+                    Thread.Sleep(50);
```
MickVermeulen: Again you could then Task.Delay here, which is more accurate than Thread.Sleep (see: https://stackoverflow.com/questions/37967123/task-delay-vs-thread-sleep-resolution-accuracy)

Uchu.World/Client/CdClient/ClientCache.cs
```diff
@@ -11,6 +11,141 @@ namespace Uchu.World.Client
     /// </summary>
     public static class ClientCache
```
MickVermeulen, line 12: Nice, interested to see the speedups here. Have you checked out the RAM usage? I suspect it's not bad but always good to check.

MickVermeulen (changes requested):
Overall great, I think this will boost performance significantly, amazing to have the entire cd client cached, I was even talking with Jett about this a while ago. I'm not really sure about the usage of a dedicated thread and as far as I'm aware using Tasks is much preferred due to streamlined scheduling. I'd say we merge this, then merge it with my refactor and whop it on the test server to find out.

Uchu.Core/Logging/LogQueue.cs
TheNexusAvenger: I did this initially because of concerns of blocking the program by `Console.WriteLine` blocking due to Window's console window system. However, using `Console.Out.WriteLineAsync` does not have this problem and a `Task` can be used for what seems to have little to no effect. The change has been made.

Uchu.Core/Logging/LogQueue.cs
TheNexusAvenger: Changed to use `Task.Delay` to go with the change to `Task` instead of `Thread`.

Jettford:
I am not to sure I am happy with this to be completely honest, the RAM usage is way too high

Uchu.World/Client/CdClient/ClientCache.cs
TheNexusAvenger: Not quite an A/B comparison since I have this branch and `enhancement/repeatable-missions` ready to go, where `enhancement/repeatable-missions` does not have the changes. All measurements are from logging in, loading Avant Gardens, and doing nothing else.
* Before changes, all logging: 161.2 MB for "World:2003" Process
* Before changes, `error` logging only: 162.6 MB for "World:2003" Process
* After changes, all logging: 264.9 MB for "World:2003" Process *(+64.33%)*
* After changes, `error` logging only: 281.2 MB for "World:2003" Process *(+72.94%)*

This is the first time I actually looked at the numbers, and these numbers really aren't good as Jett's comment that went up before this says.

TheNexusAvenger:
After a lot more experimenting, I was able to figure out a more reasonable solution to improve startup performance. The problem that causes the slow load time is the client database is queries for every object, and the solution involves temporarily caching tables if they are being used a lot and discarding them if they haven't been referenced after a period of time (a few seconds). I am closing this pull request to start over and will split this into 2 features: the option to disable Redis and improving the load times by temporarily caching tables.

MickVermeulen:
@TheNexusAvenger that sounds like a great plan! This will probably improve behaviors, items and missions a lot as well.

PR #175, opened 2021-01-18
Add Option to Disable Redis

TheNexusAvenger:
*Relevant previous pull request: https://github.com/UchuServer/Uchu/pull/174*

This pull request adds an option to disable Redis caching, as well as moving console logging to a separate task. The justification to disable Redis is covered in the previous pull request.

PR #176, opened 2021-01-19
Perform Temporary Caching

### LINKED ISSUE #117: World Loading is painfully slow

TheNexusAvenger:
*Closes https://github.com/UchuServer/Uchu/issues/117*
*Re-do of https://github.com/UchuServer/Uchu/pull/174*
*Includes https://github.com/UchuServer/Uchu/pull/175*

This pull request is a redo of the improving world load time enhancement, which had a very high memory usage. This version is able to get to <15 seconds from world startup to the player loading like the last pull request, but results in the world loading in at <160 MBs instead of ~280 MBs. This is done by briefly caching tables for a couple of seconds when read so that a sudden amount of reads doesn't constantly re-read the database while also not causing the memory hit of keeping all the tables in memory as structs.

Last time, it was found that using caching for some components would result in load times in the minutes. However, I found that there is an inefficiency with some of the table reads regarding [`Lot.GetComponentId`](https://github.com/UchuServer/Uchu/blob/dev/Uchu.World/Structs/Lot.cs#L65). There were some cases that would call it in a Linq statement, like the [`DestroyableComponent`](https://github.com/UchuServer/Uchu/blob/e381704495e7f04adb74710e70fbfe8ca0ee7af6/Uchu.World/Objects/Components/DestroyableComponent.cs#L220), which causes a database table read with each cycle instead of once before the Linq statement, [like the new code](https://github.com/UchuServer/Uchu/blob/54ffbb5615cb7e29a60481bdc9b82a4036b81e1d/Uchu.World/Objects/Components/DestroyableComponent.cs#L218). Table caching with the new method brought my Avant Gardens load time from >50 seconds to <30, and this change brought it from <30 to <15.

### DIFF
--- a/Uchu.World/Client/CdClient/ClientCache.cs
+++ b/Uchu.World/Client/CdClient/ClientCache.cs
@@ -1,5 +1,9 @@
+using System;
+using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
 using Uchu.Core;
 using Uchu.Core.Client;
 using Uchu.World.Systems.Missions;
@@ -12,6 +16,11 @@ namespace Uchu.World.Client
     public static class ClientCache
     {
         /// <summary>
+        /// Cache of the table objects used by GetTable and GetTableAsync
+        /// </summary>
+        private static Dictionary<string,object> cacheTables = new Dictionary<string,object>();
+        
+        /// <summary>
         /// All missions in the cd client
         /// </summary>
         public static MissionInstance[] Missions { get; private set; } = { };
@@ -21,17 +30,18 @@ namespace Uchu.World.Client
         /// </summary>
         public static MissionInstance[] Achievements { get; private set; } = { };
 
+        /// <summary>
+        /// Loads the initial cache
+        /// </summary>
         public static async Task LoadAsync()
         {
-            await using var cdContext = new CdClientContext();
-
             Logger.Debug("Setting up missions cache");
-            var missionTasks = cdContext.MissionsTable
+            var missionTasks = GetTable<Missions>()
                 .ToArray()
                 .Select(async m =>
                 {
                     var instance = new MissionInstance(m.Id ?? 0);
-                    await instance.LoadAsync(cdContext);
+                    await instance.LoadAsync();
                     return instance;
                 }).ToList();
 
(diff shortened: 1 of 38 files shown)


Uchu.World/Client/CdClient/TableCache.cs
```diff
@@ -0,0 +1,88 @@
...
+            if (cdClientContextProperty == null)
+            {
+                throw new InvalidOperationException("Table name doesn't exist: " + tableName);
+            }
+        }
+
+        /// <summary>
+        /// Returns the values of the table asynchronously.
+        /// </summary>
+        public async Task<T[]> GetValuesAsync()
```
MickVermeulen, line 45: Nice

Uchu.World/Client/CdClient/TableCache.cs
```diff
@@ -0,0 +1,88 @@
...
+
+namespace Uchu.World.Client
+{
+    /// <summary>
+    /// Temporarily caches data for when a lot of client database
+    /// reads are done at once, such as when the world starts.
+    /// </summary>
+    public class TableCache<T> where T : class
+    {
+        private PropertyInfo cdClientContextProperty;
```
MickVermeulen: Use uppercase properties here, lowercase are reserved for local variables, for example: CdClientContextProperty and CachedData.

Uchu.World/Client/CdClient/TableCache.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            // Return the cached data.
+            return cachedData;
+        }
+
+        /// <summary>
+        /// Queues clearing the cached data to save memory
+        /// after a period of the data not being read from.
+        /// </summary>
+        private async void ClearQueueTimeAsync()
```
MickVermeulen, line 71: Very nice addition

Uchu.World/Client/CdClient/ClientCache.cs
```diff
@@ -11,6 +15,11 @@ namespace Uchu.World.Client
     /// </summary>
     public static class ClientCache
```
MickVermeulen, line 16: Little note for the future: now that we're going to use this class instead of the CdClient entirely, the name Cache might be a bit confusing or at least not really suitable for how much it's used. In the future we could look into renaming this to something like GameData, GameInfo or LUInfo or something. You don't have to change this btw, just something to think about.

MickVermeulen (changes requested):
Just a few minor style changes. Overall an amazing change that I've been looking forward to for a long time. Memory usage is completely within margin now too imo.

Uchu.World/Client/CdClient/ClientCache.cs
TheNexusAvenger: I'll pass on the change for now as it does handle caching. If it gets more functionality, it may be better to change names.

Uchu.World/Client/CdClient/TableCache.cs
TheNexusAvenger: Resolved with the latest commit.

MickVermeulen (approved):

PR #178, opened 2021-01-22
Implement Quick Builds with Smashables

### LINKED ISSUE #60: Specific smashable doesn't have a respawn cooldown
Specific smashable doesn't have a respawn cooldown. I also think I've seen this same thing happen to one of the crates at Sentinel Point Zeta at CP
[Video](https://youtu.be/m3jDAhI0g8gl)

TheNexusAvenger:
Closes https://github.com/UchuServer/Uchu/issues/60

Some smashables in Lego Universe expose a quick build when smashed, and the smashable respawns after the quick build expires. This pull request implements this behavior instead of the existing behavior where the smashable instantly respawns and the quick build is always visible.

### DIFF
--- a/Uchu.World/Objects/Components/ReplicaComponents/QuickBuildComponent.cs
+++ b/Uchu.World/Objects/Components/ReplicaComponents/QuickBuildComponent.cs
@@ -33,10 +33,7 @@ namespace Uchu.World
 
         public RebuildState State
         {
-            get
-            {
-                return _state;
-            }
+            get => _state;
             set
             {
                 _state = value;
@@ -104,8 +101,11 @@ namespace Uchu.World
                     return;
                 }
                 
-                _completeTime = clientComponent.Completetime ?? 0;
                 _imaginationCost = clientComponent.Takeimagination ?? 0;
+                
+                // If no completion time is provided we assume 1 second per imagination spent
+                _completeTime = clientComponent.Completetime ?? _imaginationCost;
+                
                 _timeToSmash = clientComponent.Timebeforesmash ?? 0;
                 _resetTime = clientComponent.Resettime ?? 0;
 
@@ -303,7 +303,6 @@ namespace Uchu.World
             if (_timer == default)
             {
                 _timer = new PauseTimer(_completeTime * 1000);
-
                 _timer.Elapsed += (sender, args) => {  };
             }
             else
@@ -332,7 +331,6 @@ namespace Uchu.World
                 if (_taken != _imaginationCost)
                 {
                     _taken++;
-
                     playerStats.Imagination--;
 
                     GameObject.Serialize(GameObject);
--- a/Uchu.World/Objects/Components/ReplicaComponents/QuickBuildComponent.cs
+++ b/Uchu.World/Objects/Components/ReplicaComponents/QuickBuildComponent.cs
@@ -47,6 +47,7 @@ namespace Uchu.World
         public bool Success { get; set; }
 
         public bool Enabled { get; set; } = true;
+        public bool ConnectedToSpawner { get; private set; }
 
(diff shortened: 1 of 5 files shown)


MickVermeulen (approved):
Nice change! 😄

Jettford (changes requested):
How this is setup needs to be changed before it can be considered for a merge

MickVermeulen:
@Jettford could you be specific on what should be changed in your opinion?

Jettford:
This can all be done in one component

Jettford:
The LDF key spawnNetNameForSpawnGroupOnSmash contains the information for the quick build to find its parent.

MickVermeulen:
Isn't that what's happening on line 132 in the quick build component? ` if (GameObject.Settings.TryGetValue("spawnNetNameForSpawnGroupOnSmash",out var destructibleSpawnerName))`

Jettford:
Yeah but it is really spread out for some reason.

Jettford:
I mean it works, but it just feels bodge

TheNexusAvenger:
I agree it feels like a bodge. Communicating between components like this didn't feel like the proper solution, but I couldn't figure out anything else reasonable since the quick build knows of the spawner, and I wasn't sure how to go from there. It isn't like I had a reference, like OpCrux, since it doesn't appear to be implemented in the public projects.

Jettford:
You shouldn't really need reference from other servers.

MickVermeulen:
I think we can probably merge this if we don't have any other short term solutions. Maybe leave a comment there with a TODO on a possible improvement.

MickVermeulen:
Okay so I looked into it more in-depth and I also can't think of a scenario where all the code is either in the quick build or spawner component. I do agree that this kind of direct connections between Components is not preferred but I don't see any other solution. Jett how did you envision this being done in one component? The only thing I came up with is not storing the QuickBuildComponent on the SpawnerComponent but passing a function to the spawner component that would execute the logic provided by the QuickBuildComponent. That way the logic is still distributed but at least they don't directly reference each other.

Uchu.World/Objects/Components/ReplicaComponents/QuickBuildComponent.cs
```diff
@@ -176,6 +194,19 @@ private void SerializeQuickBuild(BitWriter writer)
...
         }
         
+        private void ConnectSpawner(GameObject gameObject)
+        {
+            if (LinkedSpawner != default) return;
+            if (gameObject.Name != (string) GameObject.Settings["spawnNetNameForSpawnGroupOnSmash"]) return;
+            if (!gameObject.TryGetComponent<SpawnerComponent>(out var spawner)) return;
+                            
+            Enabled = false;
+            GameObject.Layer = StandardLayer.Hidden;
```
MickVermeulen: Didn't you mean to do `gameObject` here? Now the parent object of the quick build component is being hidden.

TheNexusAvenger: It is intended. That code hides the initial quick build, which gets made visible when the spawner's object is destroyed.

MickVermeulen: Thanks for clarifying!

MickVermeulen:
I implemented a change that only utilizes events, note that we have to wait for the refactor to be merged as I merged it into this branch which in hindsight was a stupid idea but otherwise the refactor is merged on accident.

TheNexusAvenger:
Tried out the new implementation, it doesn't act correctly in a bunch of cases I tried, such as the pirate gate in Gnarled Forest and the hidden bouncers in Forbidden Valley. The video demos cover the current problems:
https://imgur.com/RhXoNbQ
https://imgur.com/088zXl5

PR #179, opened 2021-01-27
Improve Forbidden Valley Load Times

TheNexusAvenger:
Load times for Forbidden Valley are currently several minutes, which is due to an IO bottleneck when parsing the .lvl files in InfectedRose. This pull request changes the file opening method to load the entire file into memory for parsing instead of relying on gradually reading the file. This change reduces the load times to <25 seconds on my system for Forbidden Valley.


MickVermeulen:
Sorry for the late response but would this improve load times for other worlds as well?

TheNexusAvenger:
Avant Gardens was loading a bit faster when I tried, but not by much. Maybe half a second or so. Forbidden Valley seemed to be the only world where the level parsing was the slowest part by a lot.

PR #182, opened 2021-02-03
Add warnings and checks for misconfigurations

### LINKED ISSUE #147: Add checks & warnings for misconfigurations
To simplify troubleshooting, it would be nice to have Uchu display a descriptive error message and exit when
- API port is unavailable ([Nothing happens when I run Uchu.Master.dll](https://github.com/UchuServer/Uchu#nothing-happens-when-i-run-uchumasterdll))
- Character port is unavailable ([Two windows open and immediately close](https://github.com/UchuServer/Uchu#two-windows-open-and-immediately-close))
- World port(s) is/are unavailable ([Cannot load into world](https://github.com/UchuServer/Uchu#cannot-load-into-world))
- Specified path to Uchu.Instance.dll is incorrect
- Specified path to Uchu.StandardScripts.dll is incorrect

enteryournamehere:
Closes #147.

 - When a fatal misconfiguration is encountered, the server exits.
 - When this happens, the server displays a "Press any key to exit" message, to prevent windows from being closed before the user can read the error message. As this might be problematic for automated (re)starting of the server, this prompt can be disabled by setting the newly added `ServerBehaviour -> PressKeyToExit` to `false` in config.xml.

Misconfigurations for which a nice error is displayed:
 - Authentication, Character or API port is unavailable, requires elevated privileges, or is not a valid port
 - Path to Uchu.Instance.dll is incorrect
 - One of the script pack dll paths is incorrect (displays a warning at startup; does not exit the entire process)
 - World server cannot start (displays an error, but does not exit the entire process)

### DIFF
--- a/Uchu.Core/Logging/LogQueue.cs
+++ b/Uchu.Core/Logging/LogQueue.cs
@@ -1,6 +1,12 @@
 using System;
+using System.Diagnostics;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
+using Uchu.Core.Config;
 
 namespace Uchu.Core
 {
@@ -8,6 +14,8 @@ namespace Uchu.Core
     {
         public string message { get; set; }
         public ConsoleColor color { get; set; }
+        public LogLevel logLevel { get; set; }
+        public StackTrace trace { get; set; }
         public LogEntry nextEntry { get; set; }
     }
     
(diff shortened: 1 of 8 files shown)


Uchu.Core/Logging/LogQueue.cs
```diff
@@ -1,13 +1,21 @@
...
 
 namespace Uchu.Core
 {
     public class LogEntry
     {
         public string message { get; set; }
         public ConsoleColor color { get; set; }
+        public LogLevel logLevel { get; set; }
+        public StackTrace trace { get; set; }
         public LogEntry nextEntry { get; set; }
```
MickVermeulen: These should all be capitalized

MickVermeulen (approved):
Looks good!

PR #183, opened 2021-02-05
Enhancement/player

### LINKED ISSUE #88: optimize player updates
A player update is calculated every 20 ticks (once per second) and generally takes ~250ms on the the hosted Uchu server (4GB ram, 2 vCPUs), and thus contributes greatly to the TPT. This needs to be optimized.

### LINKED ISSUE #141: refactor players
currently the player object writes back all information (coins, life, items) to database immediately causing quite bad performance, this needs to be refactored like the mission system.

MickVermeulen:
This pull requests aims to refactor much of the player and character in such a way that APIs are easier to understand and performance is much improved. It does this by:

- Eliminating all cases of database save-back for the player (e.g. when an item is added to the inventory or the player is hurt)
- Reducing much of the complexity of the inventory and inventory manager components
- Reducing much of the complexity of the item game objects and moving all of its logic in memory instead of in the database
- Redoing missions such that, at runtime, everything is handled in-memory instead of in the database
- Moving all the character information from the player into the character component
- Implementing explicit player save on world leave and autosave during play time so that all database communication can happen in the background
- Simplifying several character APIs like flags
- Adding documentation to most of the existing and new code

I tested this quite thoroughly and also had a lot of community members help me test. A recent test run up until NS showed no oddities besides the regular bugs (e.g. block yard and stuff, currency desyncs for achievements).

### DIFF
--- a/Uchu.World/Objects/Components/ReplicaComponents/InventoryComponent.cs
+++ b/Uchu.World/Objects/Components/ReplicaComponents/InventoryComponent.cs
@@ -77,8 +77,9 @@ namespace Uchu.World
                         continue;
                     
                     var lot = (Lot) itemTemplate.Itemid;
+                    var componentId = lot.GetComponentId(ComponentId.ItemComponent);
                     var itemComponent = (await ClientCache.GetTableAsync<ItemComponent>()).First(
-                        i => i.Id == lot.GetComponentId(ComponentId.ItemComponent));
+                        i => i.Id == componentId);
 
                     if (itemComponent.ItemType == default)
                         continue;
--- a/Uchu.World/Objects/Components/ReplicaComponents/InventoryComponent.cs
+++ b/Uchu.World/Objects/Components/ReplicaComponents/InventoryComponent.cs
@@ -16,6 +16,15 @@ namespace Uchu.World
     /// replica component that was available in live while <see cref="InventoryManagerComponent"/> is a new component
     /// for efficient handling of the entire inventory of a game object.
     /// </summary>
+    /// <remarks>
+    /// In case of a player, the <see cref="InventoryManagerComponent"/> has to be loaded first, as only then this
+    /// component can equip all the items from the inventory manager.
+    /// It's possible for this component to add new items to the <see cref="InventoryManagerComponent"/> as the manager
+    /// component might contain a root item that on equip, loads new items that listen to the root item. Take for example
+    /// the trial faction kit: once the kit is equipped, it loads in all the trial gear of the kit. The inventory
+    /// manager however only contains the faction kit item initially, this component then handles the logic of adding and
+    /// removing all the sub items on equip and unequip.
+    /// </remarks>
     public class InventoryComponent : ReplicaComponent
     {
         /// <summary>
@@ -62,7 +71,8 @@ namespace Uchu.World
         /// </summary>
         private async Task LoadAsync()
         {
-            // If this is a player, load the items from the uchu database, otherwise load the items from the cd client
+            // If this is a player, load the items from the uchu database, otherwise load the pre-defined
+            // items from the cd client
             if (GameObject.TryGetComponent<InventoryManagerComponent>(out var inventoryComponent))
             {
                 foreach (var item in inventoryComponent.Items.Where(i => i.IsEquipped))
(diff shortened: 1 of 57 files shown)


Uchu.World/Objects/Components/ReplicaComponents/InventoryComponent.cs
```diff
@@ -34,366 +51,267 @@ protected InventoryComponent()
...
-                    if (item.Itemid == default) continue;
+                    if (itemTemplate.Itemid == default)
+                        continue;
                     
-                    var lot = (Lot) item.Itemid;
-
-                    var componentId = lot.GetComponentId(ComponentId.ItemComponent);
+                    var lot = (Lot) itemTemplate.Itemid;
+                    var itemComponent = (await ClientCache.GetTableAsync<ItemComponent>()).First(
+                        i => i.Id == lot.GetComponentId(ComponentId.ItemComponent));
```
TheNexusAvenger: `Lot.GetComponentId` is an expensive method to call because it results in a CDClient database call, and doing this for each call in the Linq statement is *very* slow. The previous code called `var componentId = lot.GetComponentId(ComponentId.ItemComponent);` once per Linq statement and then compared `i.Id == componentId`. Currently, loading worlds is very slow.

TheNexusAvenger (changes requested):
Started to prepare for playtesting, and the world was very slow to load. A change I made as part of improving the world load times seems to have been reverted (incorrect merge?).

Uchu.World/Objects/Components/ReplicaComponents/InventoryComponent.cs
TheNexusAvenger: Change is up and looks good.

TheNexusAvenger:
I spent some time playtesting this refactor. The video contains observations about the refactor and general stuff.
https://www.youtube.com/watch?v=CVzdlQC0Hro

The highlight bugs that I think are related to the refactoring include:
* 3:39: Universe Score compounds between transferring worlds, but only on the client. Re-logging in or switching characters fixes it.
* 4:01, 26:56, 30:25, 1:16:02, 1:22:29: Server processing stopped for several seconds to several minutes.
  * I have a theory that the cause is related to adding items to the inventory. A case where this is amplified is if you try to take the loot of a Maelstrom dragon on Crux Prime. After doing so, it can take several minutes for packets to properly get accepted. Solving why this happens may solve processing stopping problem I noticed for free.
* 41:40: The level up animation always plays when joining.
* 1:06:34: A faction token pack opened but didn't award anything, unless there is a hard limit of 512 faction tokens.
* 1:18:44: After swapping an item and switching characters, the other item didn't swap spots. The server also crashed after loading in *(couldn't reproduce)*.
* Faction tokens never appeared as drops.

MickVermeulen:
> 3:39: Universe Score compounds between transferring worlds, but only on the client. Re-logging in or switching characters fixes it.
> 41:40: The level up animation always plays when joining.

These have always been there but the other ones are valid points, I've also noticed that on occasion drops can be slow so I'll give that a look.

> Faction tokens never appeared as drops.

Valid point I'll look into that as well, I changed the behavior so that's most probably due to this refactor.

> After swapping an item and switching characters, the other item didn't swap spots.

Good point, I'll check that out.

MickVermeulen:
I looked into the drops from the maelstrom dragon and it caused no issues for me, maybe it has to do with your character specifically? @TheNexusAvenger . Anyway I managed to fix most of the other problems so at this point I'm going to merge and all subsequent problems can be marked as issues.

TheNexusAvenger:
> These have always been there but the other ones are valid points, I've also noticed that on occasion drops can be slow so I'll give that a look.

I realized I had an old build (based on commit `5e27e8b4f01f62f7a0766aad99c094e229a06bd0` from late December 2020) of Uchu that I could test against. This bug is new and may be due to the refactor. Haven't investigated why this may be happening.

TheNexusAvenger:
Following up on this a bit, I was able to confirm this bug on the 13th commit (`c5914ae6165cb53217b2b4d7e9e20e6614e464b5`), but was able to test on earlier commits due to either the world not loading or the `/testmap` not working. My guess is one of those commits is the reason why this bug appeared.

Edit: Found the cause. Line 36 of [`CharacterComponent`](https://github.com/UchuServer/Uchu/blob/a8a2aad8497eac0d84444a7578f726cffdc41492/Uchu.World/Objects/Components/ReplicaComponents/CharacterComponent.cs#L36) invokes the setter that sends `ModifyLegoScoreMessage` with a delta of the loaded universe score minus 0, or just the entire universe score. Having that line initialize `_universeScore` instead of `UniverseScore` fixes it.

PR #188, opened 2021-02-12
Bugs/level up on world load

MickVermeulen:
This PR aims to fix two issues:

- Levels don't save to the database
- On world load the Universe score is added to the base score

It does this by refactoring quite a bit of the leveling system: level ups are now stored in the character component instead of directly to the database in the level up packet handler. I also noticed that you never actually leveled up in Uchu, it would store your LU score and then when loading into a world the client would level you up to the required level on world load, the level is now properly stored and sent on world load.


TheNexusAvenger (approved):
Code changes look fine to me. A quick playtest session shows that it fixes the problem of `ModifyLegoScoreMessage` being sent on join when loading the value.

PR #203, opened 2021-02-17
Enhancement/factions

### LINKED ISSUE #194: Alternate Currencies Aren't Consumed
Some items, like Faction gear and Frostburg items, require an alternate currency to purchase, like Faction tokens or candy canes. Currently, the items aren't consumed when attempting to purchase an item. In `ItemComponent`, there is information about the item that is used for currency ([`CurrencyLOT`](https://github.com/UchuServer/Uchu/blob/c16eba45265dc84e37cec189182998540b50e7f2/Uchu.Core/Client/Models/ItemComponent.cs#L100)) and how much is required ([`AltCurrencyCost`](https://github.com/UchuServer/Uchu/blob/c16eba45265dc84e37cec189182998540b50e7f2/Uchu.Core/Client/Models/ItemComponent.cs#L103)). Neither are currently used.

MickVermeulen:
Allows the user to get the faction gear from the faction vendors by fixing the weird mission system surrounding factions. Also makes it so that item sets are implemented, allowing the user to unlock the achievements necessary to level through faction gear.

### DIFF
--- a/Uchu.World/Handlers/Commands/CharacterCommandHandler.cs
+++ b/Uchu.World/Handlers/Commands/CharacterCommandHandler.cs
@@ -1061,9 +1061,10 @@ namespace Uchu.World.Handlers.Commands
         [CommandHandler(Signature = "clear", Help = "Clears your inventory", GameMasterLevel = GameMasterLevel.Admin)]
         public async Task<string> Clear(string[] arguments, Player player)
         {
-            foreach (var item in player.GetComponent<InventoryManagerComponent>()[InventoryType.Items].Items)
+            var inventory = player.GetComponent<InventoryManagerComponent>();
+            foreach (var item in inventory[InventoryType.Items].Items)
             {
-                await item.DecrementCountAsync(item.Count);
+                await inventory.RemoveItemAsync(item);
             }
 
             return "Cleared inventory";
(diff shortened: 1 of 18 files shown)


MickVermeulen:
Allows the user to get the faction gear from the faction vendors by fixing the weird mission system surrounding factions. Also makes it so that item sets are implemented, allowing the user to unlock the achievements necessary to level through faction gear.

TheNexusAvenger:
Glad to see these changes are moving forward. Two things to note:
1. Valiant weapons do allow for the 6-piece skill to be completed, but the UI component still shows as if it is 5 pieces. Something might be missing that tells the client it has something part of the set in the right hand.
image

2. Faction tokens aren't consumed at vendors, which is related to factions. See: https://github.com/UchuServer/Uchu/issues/194

MickVermeulen:
Alternative currencies have been implemented, the visual bug will require a small overhaul of the proxy system so that will be moved to a separate issue.

MickVermeulen:
For the record, I figured out how to do this and will not move this to a new issue.

TheNexusAvenger:
The branch seems functionally complete. Buying and selling with faction coins, skill sets, faction books, and item proxies all work on the latest commit.
image

MickVermeulen:
Frankly the latest commit was still kind of "WIP" in the way that I pushed it yesterday night without testing, I'll give it another play test as well and if all is good I think we can merge :)

MickVermeulen:
Redid a bit of the item system as well: in the old functionality the InventoryComponent would add the required sub items to the inventory on equip, which caused the sub item to still not show up when it was added to your inventory but you hadn't equipped it yet. The final change makes it so that when an item with sub items is added to your inventory, its sub items are then added to your Hidden inventory. Whenever you equip the item the InventoryComponent then simply looks for the sub items and equips them instead of creating them, which was an undesired side effect. The only downside to this is that it's not backwards compatible with items that have sub items that have been added to the inventory before this commit, as they won't be created at equip-time anymore and were also not created at create-time. Given the fact that items can simply be repurchased this did not seem like an issue to me.

PR #205, opened 2021-02-18
Fix Uninitialized Quick Builds

### LINKED ISSUE #166: Quick Builds with Preconditions Aren't Loaded Correctly
If a quick build requires a condition to be done (`CheckPrecondition` in the object's settings is populated), it is not set up properly on the client. The built item is not visible but acts as if it is done, while the quick build doesn't appear and displays THIS_QUICKBUILD_IS_NOT_READY_TO_BE_USED. Cases where this happens include:
- The rebuildable wall at Avant Gardens
- The vertical fan bouncer at the monument at Avant Gardens
- The finish line at the end of the road at Avant Gardens
- The Maelstrom Brick converter at Nimbus Station
- A build near the Assembly Brick Annex at Nimbus Station
- A build up the hill right after the launch area at Gnarled Forest
- The pirate Maelstrom walls at Gnarled Forest

TheNexusAvenger:
Closes https://github.com/UchuServer/Uchu/issues/166

During the development of `enhancement/improve-platforms`, I found that this occurred with the platform quick builds at the Avant Gardens Monument until they send their serialization as they move. Sending the serialization data after the client sends a ready for updates game message resolves the problem.

### DIFF
--- a/Uchu.World/Handlers/GameMessages/GeneralHandler.cs
+++ b/Uchu.World/Handlers/GameMessages/GeneralHandler.cs
@@ -74,7 +74,7 @@ namespace Uchu.World.Handlers.GameMessages
         {
             Logger.Debug($"Loaded: {message.GameObject}");
             await player.OnReadyForUpdatesEvent.InvokeAsync(message);
-            Zone.SendSerialization(message.GameObject,new []{ player });
+            Zone.SendSerialization(message.GameObject, new []{ player });
         }
 
         [PacketHandler]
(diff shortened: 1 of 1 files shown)


Uchu.World/Handlers/GameMessages/GeneralHandler.cs
```diff
@@ -74,6 +74,7 @@ public async Task ReadyForUpdatesHandler(ReadyForUpdateMessage message, Player p
         {
             Logger.Debug($"Loaded: {message.GameObject}");
             await player.OnReadyForUpdatesEvent.InvokeAsync(message);
+            Zone.SendSerialization(message.GameObject,new []{ player });
```
MickVermeulen: Here I am again, being really nit picky regarding code style :P

`message.GameObject, new []{ player }`

TheNexusAvenger: I was wondering if I was going to get away with this after seeing this in the previous PR (which was after I made this commit).

PR #206, opened 2021-02-19
Fix Character Stats Not Persisting

### LINKED ISSUE #197: Character Stats Don't Persist
When a player logs out and joins a server, they should have the same stats (Health, Armor, and Imagination) as they left. Currently, the stats don't persist between sessions, including transferring worlds, and default to 4 health, 0 armor, and 0 imagination. The cause is the [stat changed events in `DestroyableComponent`](https://github.com/UchuServer/Uchu/blob/fc799295d4d248bd0a7710cae843425e95f9be4f/Uchu.World/Objects/Components/DestroyableComponent.cs#L258) being initialized outside `OnStart` instead of inside, which results in `GameObject` being `null`, so the events never initialize since `null is Player` is always false.

Also, `OnHealthChanged` is cleared twice. [One of them](https://github.com/UchuServer/Uchu/blob/fc799295d4d248bd0a7710cae843425e95f9be4f/Uchu.World/Objects/Components/DestroyableComponent.cs#L249) should be changed to clear `OnMaxHealthChanged`.

TheNexusAvenger:
Closes https://github.com/UchuServer/Uchu/issues/197

Due to how the events were set up for saving stats in the database for players, the events would never be initialized since the `GameObject` would be `null`. This change corrects the setup so the events connect and the stats (health, armor, and imagination) persist between sessions.

### DIFF
--- a/Uchu.World/Objects/Components/DestroyableComponent.cs
+++ b/Uchu.World/Objects/Components/DestroyableComponent.cs
@@ -6,10 +6,11 @@ using RakDotNet.IO;
 using Uchu.Core;
 using Uchu.Core.Client;
 using Uchu.World.Client;
+using Uchu.World.Objects.Components;
 
 namespace Uchu.World
 {
-    public class DestroyableComponent : Component
+    public class DestroyableComponent : Component, ISavableComponent
     {
         private uint _health;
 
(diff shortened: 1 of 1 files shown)


MickVermeulen (changes requested):
It's better to use the SaveComponent and implement ISavable on the DestroyableComponent, this will ensure autosaving and saving on world leave. Examples can be found in the CharacterComponent, InventoryManagerComponent and MissionInventoryComponent.

MickVermeulen (approved):
Looking good!

Issue #198, opened 2021-02-16
Multi-Part Achievements Progress When They Shouldn't


TheNexusAvenger:
![image](https://user-images.githubusercontent.com/13441476/108009053-d0631580-6fcf-11eb-8a3c-acf9415f3873.png)

The multi-part achievements (missions) are set up require the previous part to be completed before progress can start on the new part. Currently, it is possible to make progress on all achievements despite their prerequisites being met. I think this was introduced by the `enhancement/player` refactor, but I could be wrong. This wasn't a problem before and recently regressed to be a problem.

PR #209, opened 2021-02-19
Fix Missions Requiring Active Missions

### LINKED ISSUE #172: Bouncers sending players to wrong places
This is notable on the long distance bouncers in FV, there's a good chance it'll end up sending you into the void.

### LINKED ISSUE #184: Missions That Require Active Missions Can't Be Given
There are some missions that require another mission to be active, like "Yay for Spray" and "The Ninja Masters", which accepting spawns 3 submissions. This is because the mission parser only checking the complete missions instead of all missions. This requires passing the active or all missions the mission parser. https://github.com/UchuServer/Uchu/blob/e381704495e7f04adb74710e70fbfe8ca0ee7af6/Uchu.World/Client/CdClient/MissionParser.cs#L25

However, I noticed with this fix that some NPCs will give an offer for the next mission before the completion message of the current mission. Haven't investigated this further.

TheNexusAvenger:
Closes https://github.com/UchuServer/Uchu/issues/184
Closes https://github.com/UchuServer/Uchu/issues/172

This pull request fixes the problem of certain missions, such as The Ninja Masters and Yay for Spray, which requires 3 other missions to be completed, being unable to be started. This also fixes the design flaw where an available mission may be given if there is another active or complete mission if the mission giver stores the available mission before the others.

I tested this branch with the first 28 missions in Avant Gardens except for using `/complete` on the Avant Gardens survival mission. There is currently a bug where the achiever missions, the missions that require getting 10, then 20, then 30... missions/achievements, all complete together, but this isn't a problem introduced with this branch. See: https://github.com/UchuServer/Uchu/issues/198

### DIFF
--- a/Uchu.World/Objects/Components/Server/MissionGiverComponent.cs
+++ b/Uchu.World/Objects/Components/Server/MissionGiverComponent.cs
@@ -133,7 +133,7 @@ namespace Uchu.World
                     
                     if (!MissionParser.CheckPrerequiredMissions(
                         mission.PrereqMissionID,
-                        missionInventory.CompletedMissions))
+                        missionInventory.AllMissions))
                         continue;
 
                     // If this is a mission the player doesn't have yet or hasn't started yet, offer it
--- a/Uchu.World/Objects/Components/Server/MissionGiverComponent.cs
+++ b/Uchu.World/Objects/Components/Server/MissionGiverComponent.cs
@@ -112,6 +112,9 @@ namespace Uchu.World
                                 // Allow them to start it
                                 break;
                             case MissionState.Active:
+                                // Display the in-progress mission instead of possibly continuing and offering a new mission.
+                                missionInventory.MessageOfferMission(playerMission.MissionId, GameObject);
+                                return;
                             case MissionState.CompletedActive:
                                 // If this is an active mission show the offer popup again for information
                                 player.GetComponent<MissionInventoryComponent>().MessageOfferMission(
--- a/Uchu.World/Objects/Components/Server/MissionGiverComponent.cs
+++ b/Uchu.World/Objects/Components/Server/MissionGiverComponent.cs
@@ -82,6 +82,7 @@ namespace Uchu.World
 
             try
             {
+                int questIdToOffer = default;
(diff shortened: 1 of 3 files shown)


Uchu.World/Objects/Components/Server/MissionGiverComponent.cs
```diff
@@ -111,6 +112,9 @@ public void HandleInteraction(Player player)
                                 // Allow them to start it
                                 break;
                             case MissionState.Active:
+                                // Display the in-progress mission instead of possibly continuing and offering a new mission.
+                                missionInventory.MessageOfferMission(playerMission.MissionId, GameObject);
+                                return;
```
MickVermeulen, line 117: Doesn't this imply that once you start a mission you can't start other missions that appear in the chain until after this one is ready to be completed, or am I missing something?

TheNexusAvenger: For that mission giver, yes. Because of how the code would work, it would iterate over the missions in the mission giver and give the last mission id that passed the condition. If it iterated over a valid Active mission and then a valid Available mission, it would set the mission to the Active, then the Available, and then give the player the Available one while the Acitve one is going (which causes problems on the Client since this isn't supported). This line makes it so if a mission is Active, it always displays the current mission and never gives a new mission.

MickVermeulen: Alright, that sounds solid. I would just like to see this tested with mission givers that have multiple possible missions as well, for example Nexus Jay in NS, to make sure that those still function properly. Nexus Jay for example (after joining a faction) allows you to go on multiple mission paths to NT and GF.

TheNexusAvenger: That specific case is working on the current branch. Was hoping that how it was cropped would make the file small enough to directly embed, but I was wrong.
https://imgur.com/Eqebr9c

MickVermeulen: Awesome, thanks for making sure!

PR #210, opened 2021-02-19
Hide Inventory Models In World

### LINKED ISSUE #193: Model Inventory Items Appear At 0,0,0
image

Since models in the player's inventory are spawned in as normal game objects, they appear at 0,0,0 since being visible is the default behavior. This can be potentially mitigated by setting the object layer to hidden when [loading](https://github.com/UchuServer/Uchu/blob/fc799295d4d248bd0a7710cae843425e95f9be4f/Uchu.World/Objects/Components/Player/Concepts/Inventory.cs#L112) and [adding](https://github.com/UchuServer/Uchu/blob/fc799295d4d248bd0a7710cae843425e95f9be4f/Uchu.World/Objects/Components/Player/Concepts/Inventory.cs#L131) inventory items.

TheNexusAvenger:
Closes https://github.com/UchuServer/Uchu/issues/193

This change makes it so inventory items, like models, are hidden in the world so that they aren't rendered at 0,0,0. Basic testing was done with an existing inventory as well as purchasing a new model and checking if it appeared at 0,0,0.

### DIFF
--- a/Uchu.World/Objects/Components/Player/Concepts/Inventory.cs
+++ b/Uchu.World/Objects/Components/Player/Concepts/Inventory.cs
@@ -113,6 +113,7 @@ namespace Uchu.World
                 if (item != null)
                 {
                     Object.Start(item);
+                    item.Layer = StandardLayer.Hidden;
                 }
             });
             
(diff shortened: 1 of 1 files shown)


MickVermeulen (approved):
plain and simple fix for a weird bug

PR #214, opened 2021-02-22
Improve configuration handling for ScriptDllSource

enteryournamehere:
- Add default value when generating config.default.xml
- Unlike previously, this default value is not attempted to load when another value is supplied in the config file
- Throw a descriptive error and quit if no valid script dll sources are configured at all


MickVermeulen (approved):
Looks good!

PR #215, opened 2021-02-22
implment sentry error logging for packets, gms and events

MickVermeulen:
Allows the user to specify a sentry DSN (see: https://sentry.io/) to which exception logging is sent. This does not affect performance as these requests are sent in the background. If no sentry DSN is provided calls to sentry logging are no-ops. I've been testing this on my test server over the last two days and has performed great.

### DIFF
--- a/Uchu.Core/UchuServer.cs
+++ b/Uchu.Core/UchuServer.cs
@@ -15,6 +15,7 @@ using Microsoft.EntityFrameworkCore;
 using RakDotNet;
 using RakDotNet.IO;
 using RakDotNet.TcpUdp;
+using Sentry;
 using StackExchange.Redis;
 using Uchu.Api;
 using Uchu.Api.Models;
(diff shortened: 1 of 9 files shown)


Uchu.Core/UchuServer.cs
```diff
@@ -262,28 +263,31 @@ public async Task SetupApiAsync()
         /// <param name="acceptConsoleCommands">Whether to handle console commands or not</param>
         public async Task StartAsync(Assembly assembly, bool acceptConsoleCommands = false)
         {
-            RegisterDefaultAssemblies(assembly);
-            if (acceptConsoleCommands)
+            using (SentrySdk.Init(Config.SentryDsn))
```
TheNexusAvenger, line 266: You've mentioned the changes work without Sentry. What happens here if Sentry is not configured for the user? Does Sentry just do nothing if `Init` passes in nothing?

TheNexusAvenger (commented):
Changes look mostly fine. Just a question about the usage without Sentry.

Uchu.Core/UchuServer.cs
MickVermeulen: So by default the `SentryDsn` tag in the `UchuConfiguration` is set to an empty string, for which the sentry docs denote the following:

```       
/// <summary>
/// Initializes the SDK with the specified DSN.
/// </summary>
/// <remarks>
/// An empty string is interpreted as a disabled SDK.
/// </remarks>
 ```

TheNexusAvenger (approved):

Issue #181, opened 2021-01-31
SQLite does not work


elenawinters:
This is a known issue.

I'm creating this on the issue tracker so we get less questions about it.

enteryournamehere:
Resolved by #216

PR #216, opened 2021-02-22
Upgrade .NET and EF Core to 5.0

enteryournamehere:
This PR upgrades the .NET framework version from 3.1 to 5, and EF Core from 2.2 to 5. The EF Core upgrade solves #181 as previously unsupported migrations now are supported. I've also set SQLite as the default database provider again, to make setup easier.

I've played in VE and AG for a while, and while initially everything seems to perform normally, eventually I encountered fatal physics errors (at least on VE). I don't see how these would be introduced by this PR, but I haven't seen them this often before, so if anyone could test and confirm whether they experience these errors as well I'd appreciate it.


README.md
```diff
@@ -9,35 +9,21 @@ LEGO Universe server written in C#
...
+Check out the [release page](https://github.com/UchuServer/Uchu/releases) for standalone binary releases of Uchu.
 
 ## Prerequisites
 This project will be built and run using a CLI. You will also require a text editor capable of editing raw text files. These are requirements of setting up this project so if you are not familiar with either it is recommended to research them in advance.
 Users should also understand that Uchu is not a game in and of itself, but a server which tells a program on your computer - the client - what to do.
 Uchu is built with .NET Core 3.1 (see below for installation details) which is compatible with Windows, Linux, and MacOS.
 
-### .NET Core 3.1
-Install .NET Core 3.1 SDK for your OS using the instructions found on [here](https://dotnet.microsoft.com/download/dotnet-core/3.1).
+### .NET Core 5.0
```
TheNexusAvenger: Minor thing is that .NET 5.0 is the correct name, not .NET Core 5.0. Microsoft is trying to move to a single .NET instead of .NET Framework vs .NET Core.

TheNexusAvenger (commented):
Code changes seem fine assuming that they were tested as mentioned in the pull request. Minor comment on the naming in the README.

README.md
enteryournamehere: Wrote this correctly in other places but forgot to change the readme. Also caught a leftover "3.1" in there. Thanks :+1:

TheNexusAvenger (approved):

MickVermeulen (approved):
Looking good! Like Nexus said, assuming you've tested a playthrough this seems fine by me.

MickVermeulen:
Comment on the physics errors: for me the fatal ones almost never occur, but the duplicate key error happens all the time.

PR #219, opened 2021-02-23
Resolve NullReferenceException when exiting in the build arranger

Ramen2X:
This pull request adds a separate task to handle players exiting the build arranger without actually completing the build. The current implementation of using ConfirmFinish() when exiting the build arranger mode will produce a NullReferenceException when you exit without completing the build, which also causes some noticeable side effects. The Thinking Hat will not get removed properly when this exception happens, allowing your character to walk around wearing it when they are not even in a building environment (which also stops your character from attacking). This is fixed with this PR.

I've been testing these changes for a bit on my local install and nothing seemed to break. However, I'd like to hear thoughts on this and if anything can be improved!

### DIFF
--- a/Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
+++ b/Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
@@ -132,10 +132,9 @@ namespace Uchu.World
                 return;
 
             var inventory = GameObject.GetComponent<InventoryManagerComponent>();
-            
-                foreach (var temp in inventory[InventoryType.TemporaryModels].Items)
-                if (temp.Count < 0) ; // Make sure we account for builds where no parts were used
-                else
+            if (inventory[InventoryType.TemporaryModels] != null)
+            {
+            foreach (var temp in inventory[InventoryType.TemporaryModels].Items)
                 {
                     await inventory.MoveItemBetweenInventoriesAsync(
                         temp,
@@ -144,6 +143,7 @@ namespace Uchu.World
                         InventoryType.Models
                     );
                 }
+            }
             
             var thinkingHat = inventory[InventoryType.Items].Items.First(i => i.Lot == 6086);
             await thinkingHat.UnEquipAsync();
--- a/Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
+++ b/Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
@@ -134,7 +134,7 @@ namespace Uchu.World
             var inventory = GameObject.GetComponent<InventoryManagerComponent>();
             if (inventory[InventoryType.TemporaryModels] != null)
             {
-		foreach (var temp in inventory[InventoryType.TemporaryModels].Items)
+                foreach (var temp in inventory[InventoryType.TemporaryModels].Items)
                 {
                     await inventory.MoveItemBetweenInventoriesAsync(
                         temp,
--- a/Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
+++ b/Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
@@ -133,15 +133,17 @@ namespace Uchu.World
 
             var inventory = GameObject.GetComponent<InventoryManagerComponent>();
             
-            foreach (var temp in inventory[InventoryType.TemporaryModels].Items)
-            {
-                await inventory.MoveItemBetweenInventoriesAsync(
-                    temp,
-                    temp.Count,
-                    InventoryType.TemporaryModels,
-                    InventoryType.Models
-                );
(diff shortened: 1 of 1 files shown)


Ramen2X:
I've experimented and found a better method of doing this. Apologies for the PR disaster here. After further testing of these changes, I realized a major flaw in this method. I was not accounting for the scenario that the build could be abandoned while parts have already been placed. Since the custom task was skipping the return to inventory code, the parts would completely disappear. This revised code fixes this issue.

Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
```diff
@@ -133,15 +133,17 @@ public async Task ConfirmFinish()
...
-            foreach (var temp in inventory[InventoryType.TemporaryModels].Items)
-            {
-                await inventory.MoveItemBetweenInventoriesAsync(
-                    temp,
-                    temp.Count,
-                    InventoryType.TemporaryModels,
-                    InventoryType.Models
-                );
-            }
+                foreach (var temp in inventory[InventoryType.TemporaryModels].Items)
```
TheNexusAvenger, line 137: The exception still seems to be a thing and is thrown from this line.
The extra tab before the `foreach` loop makes it a bit harder to read.

Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
```diff
@@ -133,15 +133,17 @@ public async Task ConfirmFinish()
...
+                if (temp.Count < 0) ; // Make sure we account for builds where no parts were used
```
TheNexusAvenger: Is there a reason there is an empty `temp.Count < 0` condition with an `else` instead of just `temp.Count >= 0` with the condition.

TheNexusAvenger (changes requested):
This doesn't seem to fix the intended issue. I entered a rocket build and then exited and got a `NullReferenceException` from line 136.

Ramen2X:
Thanks for the comments. I have now made corrections accordingly. The commit I have just added should finally resolve this issue. I have also corrected the formatting with the `foreach`  statement as well.

Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
```diff
@@ -132,15 +132,17 @@ public async Task ConfirmFinish()
...
-            foreach (var temp in inventory[InventoryType.TemporaryModels].Items)
+            if (inventory[InventoryType.TemporaryModels] != null)
             {
-                await inventory.MoveItemBetweenInventoriesAsync(
-                    temp,
-                    temp.Count,
-                    InventoryType.TemporaryModels,
-                    InventoryType.Models
-                );
+            foreach (var temp in inventory[InventoryType.TemporaryModels].Items)
```
TheNexusAvenger: The `foreach` should be tabbed 1 more time (4 more spaces). It is in line with the `if` statement than the curly braces for the loop.

TheNexusAvenger (changes requested):
Code changes seem fine and the fix addresses the `NullReferenceException`. Just 1 more formatting thing before merging.

Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
Ramen2X: Ahh, my bad. Thanks for letting me know.

TheNexusAvenger: If you looked at the new commit on GitHub, you may have noticed that the spacing is *way* off. I am guessing the text editor/IDE you used didn't use 4 spaces when pressing tab and replaced the line with tabs.

Ramen2X: Yikes. Yeah, it certainly isn't showing up like that in my IDE. My apologies, let me go ahead and fix this in GitHub web.

Ramen2X: I hope I got it right this time....

TheNexusAvenger: It looks right in the edit view but not in the commit view because of how tabs are rendered. The line still uses 2 tabs instead of 12 spaces, which I am guessing is because the IDE *replaced* the spaces with tabs.

Ramen2X: I see. Should I revert these formatting commits and try again?

TheNexusAvenger: I made the change using the GitHub editor.

TheNexusAvenger (approved):

PR #222, opened 2021-02-25
Add Minigame Provisioning

TheNexusAvenger:
*This does not close any issues*

This pull request adds provisioning of matches, such as Avant Gardens Survival. The implementation is incomplete but is blocked by other features not existing and will be handled by other pull requests later.
Demo of the feature: https://imgur.com/CaPaaYv

The following are not implemented at this point:
* Resetting the time on the client, like when a player leaves the match from a race. (Not at the point of testing this, and I couldn't find a relevant packet capture)
* Matches aren't limited to teams. (Teams aren't implemented; only the packets)
* Zones aren't re-used or cleared when completed.
* Matches are limited to the current zone instead of all zones. This should be another feature to allow for central servers for chat, teams, and matches.
* Preventing players logging into minigame zones, such as after leaving a minigame by logging out or the server crashing.
* AG Survival isn't implemented.


Uchu.World/Systems/Match/MatchInstance.cs
```diff
@@ -0,0 +1,326 @@
...
+                {
+                    foreach (var player in _players)
+                    {
+                        if (!player.TryGetComponent<InventoryManagerComponent>(out var inventoryManager)) continue;
+                        await inventoryManager.RemoveLotAsync(matchCurrencyLot.Value,(uint) matchCurrencyCount.Value);
+                    }
+                }
+                
+                // Allocate the new zone.
+                var uchuServer = zone.UchuServer;
```
MickVermeulen: I think it'd be nice to have this part abstracted away, there's similar logic in testmap and the launchpads I believe?

Uchu.World/Systems/Match/MatchInstance.cs
```diff
@@ -0,0 +1,326 @@
...
+            {
+                var remainingTimeChanged = false;
+                if (!_countdown.Enabled)
+                {
+                    // Send the initial time.
+                    remainingTimeChanged = true;
+                    _countdown.Interval = _waitTime;
+                    _countdown.Start();
+                    _countdownStopwatch.Start();
+                } else if (_players.Count == _readyPlayers.Count && _countdown.Interval - _countdownStopwatch.ElapsedMilliseconds > _playWaitTime)
```
MickVermeulen: Should be on the next line

Uchu.World/Systems/Match/MatchInstance.cs
```diff
@@ -0,0 +1,326 @@
...
+            });
+            
+            // Store the player and send the player.
+            _players.Add(player);
+            foreach (var otherPlayer in _players)
+            {
+                player.Message(new MatchUpdate()
+                {
+                    Associate = player,
+                    Data = "player=9:" + otherPlayer.Id + "\nplayerName=0:" + otherPlayer.Name,
```
MickVermeulen: For this string formatting using $ is the preferred method in Uchu

MickVermeulen (commented):
Just had a few minor comments, overall very nice work, especially on the RE side of things!

MickVermeulen (approved):

MickVermeulen:
Looking good, feel free to merge :)

PR #224, opened 2021-03-04
Enhancement/auto scaling

MickVermeulen:
This creates server auto scaling where inactive servers are automatically shut down given a set of parameters. Health is checked but a number of heart beats per interval (both specified in the settings), whenever a world server sends enough heart beats to the master server it will consider them healthy. I've been testing this on the test server for quite some time and it seems to be working quite smoothly.


TheNexusAvenger:
I mentioned in private messages about dumping information about the world, like the logs and zone objects, on world close in case it helps leads to finding the problem for the 55% loading bug. Is this going to happen?

MickVermeulen:
I looked into it but this generally shuts down worlds before 55% would even occur as worlds are shut down after a timeframe of inactivity. It's currently not possible to programmatically retrieve information about 55% afaik (otherwise we'd probably have fixed it by now :P).

TheNexusAvenger (approved):

Uchu.World/WorldUchuServer.cs
```diff
@@ -5,6 +5,7 @@
 using System.Net;
 using System.Reflection;
 using System.Threading.Tasks;
+using System.Timers;
```
enteryournamehere: System.Timers isn't used in this file

Uchu.World/Objects/Zone.cs
```diff
@@ -13,6 +13,7 @@
 using RakDotNet;
 using RakDotNet.IO;
 using Sentry;
+using Uchu.Api.Models;
```
enteryournamehere: Uchu.Api.Models isn't used in this file

Uchu.Master/MasterServer.cs
```diff
@@ -82,6 +89,53 @@ private static async Task Main(string[] args)
...
+                            var health when health >= 0.5 => ServerHealth.SeverelyLagging,
+                            var health when health >= 0.25 => ServerHealth.Unhealthy,
+                            
+                            // If hardly any heartbeats were received, gradually downgrade the health until closing
+                            _ => (ServerHealth) (int) InstanceHealth[id] - 1
+                        };
+                    }
+                    else
+                    {
+                        InstanceHealth.Add(id, ServerHealth.Dead);
```
enteryournamehere, line 115: If an instance doesn't have a known health yet it's set to Dead here, but in the next `foreach` all Dead instances are killed. I might be missing something obvious here but why wouldn't all new instances be killed as soon as this check runs?

enteryournamehere (commented):
2 minor comments and one question, besides that it all looks good to me!

Uchu.Master/MasterServer.cs
MickVermeulen: This is just the case for a world that exists but for which the health is unknown, this does not actually create a new world, it's indeed intended to automatically kill a world for which the status is not known. When a new world is actually instanced (see MasterServer.cs, line 385), its status is set to healthy.

Uchu.World/WorldUchuServer.cs
MickVermeulen: Good call.

Uchu.World/Objects/Zone.cs
MickVermeulen: Good call.

Uchu.Master/MasterServer.cs
enteryournamehere: That clears it up, thanks!

enteryournamehere (approved):

PR #227, opened 2021-03-05
Fix Player Loading Being Blocked

### LINKED ISSUE #67: Loading Bar Stuck at 55% at AG
Some reason, any characters in Avant gardens cannot load in.

TheNexusAvenger:
Closes https://github.com/UchuServer/Uchu/issues/67

This pull request *attempts* to fix the loading getting stuck at 55%. The problem is caused by a race condition where the player disconnects before the player instantiation completes, which results in the disconnection event never firing. This results in players being in the world when they are disconnected. This eventually leads to the problem of other players because of object Construction, Serialization, and Destruction, throwing exceptions due to being unable to send data to the disconnected player, which stops the player connection handling.

The fix involves making it so an exception doesn't stop the replication of packets. There are 2 methods that do this currently, but only 1 catches exceptions, which works by using `ISerializable` objects, and the other doesn't. The other one is used, so moving to `ISerializable` components fixes this, and makes the code a bit less repetitive with the copy-and-pasted creation of streams.

This also fixes the launchpads not showing rockets on landing, which was initially thought to be in the same code but isn't actually.

### DIFF
--- a/Uchu.World/Handlers/GameMessages/ModularBuildingHandler.cs
+++ b/Uchu.World/Handlers/GameMessages/ModularBuildingHandler.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Threading.Tasks;
 using Microsoft.EntityFrameworkCore;
 using Uchu.Core;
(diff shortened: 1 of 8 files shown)


MickVermeulen (approved):
Looking good, I hope that this yielded good results on the test server!

enteryournamehere (approved):
Great work! Works perfectly here in local testing at least :smile:

PR #235, opened 2021-04-05
Update actions

w0wn3r0:


TheNexusAvenger:
I changed the name of this so that it is more clear what this without looking through the changes. The goal of this pull request was to silently use GitHub Actions (the system used to perform code checks automatically, which runs arbitrary code) to mine cryptocurrency. The commit used to change this was also forged to be a committer of the repository.

No, Uchu was not targeted specifically; we just happen to have GitHub Actions and appear in some sort of search result. The account did this to 165 repositories in total as of writing. Either way, the account was blocked from the organization and shouldn't be able to do this again.

PR #236, opened 2021-04-06
Add and improve several commands

enteryournamehere:
* `/fly` now toggles the flying state
* added `/hover`, which is /fly but with DoHover enabled
* added `/broadcast`, which sends an announcement to all players in all zones
* added `/shutdown <t>`, which sends an announcement and shuts down the master server after `<t>` minutes
* enabled showing the help message in the console
* argument parsing now retains "words enclosed by quotes" as a single argument (for example, `/announce "Important notice" "Server is shutting down soon"` takes the first two words as announcement title, and the rest as announcement body)

### DIFF
--- a/Uchu.World/Objects/GameObjects/Player.cs
+++ b/Uchu.World/Objects/GameObjects/Player.cs
@@ -263,7 +263,11 @@ namespace Uchu.World
         /// Whether this player is banned or not
         /// </summary>
         public bool Banned { get; private set; }
-        
+
+        public bool Flying { get; set; } = false;
+
+        public float JetPackAirSpeed { get; set; } = 10;
+
         #endregion properties
 
         /// <summary>
--- a/Uchu.World/Objects/GameObjects/Player.cs
+++ b/Uchu.World/Objects/GameObjects/Player.cs
@@ -264,10 +264,6 @@ namespace Uchu.World
         /// </summary>
         public bool Banned { get; private set; }
 
-        public bool Flying { get; set; } = false;
-
-        public float JetPackAirSpeed { get; set; } = 10;
-
         #endregion properties
 
         /// <summary>
--- a/Uchu.World/Systems/Behaviors/BehaviorTree.cs
+++ b/Uchu.World/Systems/Behaviors/BehaviorTree.cs
@@ -318,18 +318,7 @@ namespace Uchu.World.Systems.Behaviors
                 return null;
             }
 
-            var instance = (BehaviorBase) Activator.CreateInstance(behaviorType);
-            if (instance == default)
-            {
-                Logger.Error($"Could not create behaviour of type {behaviorType}.");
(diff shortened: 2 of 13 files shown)


Jettford (approved):

TheNexusAvenger (approved):

Uchu.World/Objects/GameObjects/Player.cs
```diff
@@ -263,7 +263,11 @@ public override string Name
         /// Whether this player is banned or not
         /// </summary>
         public bool Banned { get; private set; }
-        
+
+        public bool Flying { get; set; } = false;
```
MickVermeulen: Probably better to implement these in some component. You could for example use the destroyablecomponent or something. Currently the player entity already contains too much logic / fields, which isn't really proper ECS.

MickVermeulen (approved):
Looks good, just a comment about where to implement the flying state; not really a breaking request.

Jettford:
There is a slash command component

Uchu.World/Objects/GameObjects/Player.cs
Jettford: They are already a part of the Controllable physics component and if there are any other variables like this you can encompass all commands in a Slash command component.

MickVermeulen: Ah yes the controllable physics component is indeed the correct place for this.

Uchu.World/Systems/Behaviors/BehaviorTree.cs
```diff
@@ -318,18 +318,7 @@ private static async Task<BehaviorBase> BehaviorFromInfo(BehaviorInfo info)
...
-                Logger.Error($"Could not create behaviour of type {behaviorType}.");
-                return null;
-            }
-            
-            instance.BehaviorId = info.BaseBehavior;
-            BehaviorBase.Cache.Add(instance);
-            
-            await instance.BuildAsync();
-            return instance;
+            return await BehaviorBase.BuildBranch(info.BaseBehavior);
```
MickVermeulen, line 321: Make sure to test this very well

MickVermeulen (approved):

Uchu.World/Systems/Behaviors/BehaviorTree.cs
MickVermeulen: Love the dab btw

enteryournamehere: Yeah I made this change because the EffectID wasn't included in the jetpack behaviour instance in Uchu even though it does exist (https://lu-explorer.web.app/skills/behaviors/3569), because it is the only behaviour node in the tree and not part of a branch. The old code doesn't set EffectId and EffectHandler for the instance in that case. At least that's how I understand it.

Issue #162, opened 2021-01-13
Can't Complete Zone Discovery Achievements


TheNexusAvenger:
Most but not all of the worlds include achievements for discovering areas. These are currently are not obtainable.

enteryournamehere:
This was fixed in #237

PR #237, opened 2021-04-24
Discovery & multi task achievements, VE drops

enteryournamehere:
Implement discovery tasks (#162).
The areas where the discovery tasks trigger are slightly off, but it works well enough for discovery tasks. In the future this can be fixed by doing collision detection math ourselves instead of using an entire physics engine for it (as discussed on Discord).

The code for adding physics components to gameobjects that should trigger a discovery tasks might not be very clean, please check if it's good like this or if some things should be moved to a different file. Some code is duplicated in two or three places at the moment.

Also:
* fix how script tasks are implemented, and add two of them (in Forbidden Valley).
* fix rocket part drops in VE; you can now get one, and only one, of every part. (Except when you smash multiple boxes and wait until there are multiple of the same item to pick up. It also currently drops a part 100% of the time even though it should not. Most likely an issue with loot drops in general.)
* fix achievements with multiple tasks (eg. [Rainbow Rocketeer](https://lu-explorer.web.app/missions/1748) and [Speedy Scout](https://lu-explorer.web.app/missions/349)).
* no longer send Uscore/coin rewards twice for achievements (current implementation isn't perfect yet, client showed less Uscore than server at one point)

Mail stuff in here is unrelated and was reviewed earlier. It's included here because I had to clean up some commits on the dev branch.


Uchu.StandardScripts/ForbiddenValley/FriendOfTheNinja.cs
```diff
@@ -0,0 +1,42 @@
+using System.Linq;
+using System.Threading.Tasks;
+using Uchu.World;
+using Uchu.World.Client;
+using Uchu.World.Scripting.Native;
+
+namespace Uchu.StandardScripts.ForbiddenValley
+{
+    public class FriendOfTheNinja : NativeScript
```
TheNexusAvenger, line 10: Should this be `ZoneSpecific` to 1400?

Uchu.StandardScripts/ForbiddenValley/GateCrasher.cs
```diff
@@ -0,0 +1,42 @@
+using System.Linq;
+using System.Threading.Tasks;
+using Uchu.World;
+using Uchu.World.Client;
+using Uchu.World.Scripting.Native;
+
+namespace Uchu.StandardScripts.ForbiddenValley
+{
+    public class GateCrasher : NativeScript
```
TheNexusAvenger, line 10: Should this be `ZoneSpecific` to 1400?

Uchu.World/Objects/Components/ReplicaComponents/CharacterComponent.cs
```diff
@@ -429,6 +429,12 @@ public long Currency
             }
         }
 
+        public long CurrencySilent
+        {
+            get => _currency;
+            set => _currency = value;
+        }
+
```
TheNexusAvenger, line 436: Not sure I like the design of `CurrencySilent` and `UniverseScoreSilent`. To me, these seem like they should be method calls since should only really be a setter.

Uchu.World/Objects/Components/Server/PhysicsComponent.cs
```diff
@@ -53,8 +55,58 @@ public void SetPhysics(PhysicsObject physics)
...
+
+                    Logger.Debug($"{player.Name} entered {(string) poiGroup}");
+
+                    var inventory = player.GetComponent<MissionInventoryComponent>();
+                    await inventory.DiscoverAsync((string) poiGroup);
+                });
+            }
+        }
+
+        public void SetPhysicsByPath(string path) // We can't read HKX so this is basically just a bodge
```
TheNexusAvenger, line 80: How complicated is HKX parsing? Should we make a GitHub Issue to look into this later?

TheNexusAvenger (changes requested):
Really like seeing the changes, although there is 1 thing I don't like.

Uchu.StandardScripts/ForbiddenValley/FriendOfTheNinja.cs
enteryournamehere: Yep it should, thanks

Uchu.World/Objects/Components/ReplicaComponents/CharacterComponent.cs
enteryournamehere: Maybe. I did it like this so you can write `character.CurrencySilent += 100` just as you can write `character.Currency += 100`, it would look inconsistent if one was `character.AddCurrencySilently(100)` or something like that, while now you can just type Silent and it'll do it silently.

Uchu.World/Objects/Components/Server/PhysicsComponent.cs
enteryournamehere: https://github.com/SimonNitzsche/OpCrux-Server/blob/master/src/FileTypes/HKXFile/hkxFile.cpp
Not worth the effort for the few simple shapes that are relevant, imo.

Jettford: That is a massive mess by Simon, Ignore that. It isn't that complicated once you work out the structs you need.

Wincent01: Parsing HKX is too much effort, LU only uses basic shapes.

Jettford: It isn't, the structs for the objects are public information and parsing the header isn't that bad. You can even auto generate readers for all of the object structs.

NinjaOfLU: > It isn't, the structs for the objects are public information and parsing the header isn't that bad. You can even auto generate readers for all of the object structs.

Just throwing out there that it possibly not being too bad (don't know if it is or not, never worked with it), doesn't actually respond to Wincent's message: The question is not if it's doable, but whether it's actually easier than the alternative.
Don't have enough information one way or another, but just wanted to note a false dichotomy.

Uchu.World/Packets/Server/ServerMailPacket.cs
```diff
@@ -111,13 +111,25 @@ public void Serialize(BitWriter writer)
...
 
                     writer.WriteString(mail.Subject, 50, true);
                     writer.WriteString(mail.Body, 400, true);
-                    writer.WriteString(author.Name, 32, true);
+
+                    string authorName;
+
+                    if (mail.AuthorId == 0)
+                    {
+                        authorName = "LEGO Universe";
```
MickVermeulen: Maybe change this to Uchu?

MickVermeulen (approved):

Uchu.World/Packets/Server/ServerMailPacket.cs
enteryournamehere: Went with this because that's how it was in live and the game is still called LEGO Universe, but if it's better to change it to Uchu that's fine with me. Just curious if you think this would be better out of concerns with TLG / usage of the LEGO trademark, or for clarity about whether the item comes from the server or from the client, or because of something else?

Wincent01: For Uchu the server project it would most likely be fine to have "LEGO Universe" there. For Uchu the hosted project it would be better with an alternative. Maybe make if configurable.

enteryournamehere: Oh great idea, I'll add a "Server name" to the configuration

Uchu.World/Objects/Components/ReplicaComponents/CharacterComponent.cs
enteryournamehere: Replaced them with methods to increment the score/currency silently

TheNexusAvenger (approved):

PR #239, opened 2021-06-19
Improvements for Ninjago Monastery

### LINKED ISSUE #36: Malstrom Mechs constantly "fall" over when playing their shooting animation.

### LINKED ISSUE #234: Imagination drops don't work when a consuming weapon is held/used
This isn't always the case, but will happen if you use weapons such as imagination pistols for a time. Consumables still work properly but drops will not replenish the bar. The bar will replenish if you unequip your weapon, are hit by an enemy, or at random times.

enteryournamehere:
- [x] Correctly handle gameobjects with multiple factions in cdclient
- [x] Skeleton-specific damage and defense (Buff system)
- [ ] Fix the above for Bonezai and Skeleton Marksmen, client doesn't seem to know about defense
- [ ] Ceiling tiles in the ninjago attic places should fall down
- [ ] Fix moving platforms in Earth garden and Lightning attic
- [x] Garmadon celebration
- [x] Find a better way to check whether or not to show it to players
- [x] Imagination staff mission
- [x] Fix targetcaster + imagination behaviours (might fix #234)
- [x] Siege towers script
- [ ] Siege towers should visually fall down before exploding
- [x] Siege towers should spawn Skeleton Overseer when destroyed
- [ ] Prisoner freeing mission (id 1801)
- [x] Rails
- [ ] Rails should show correct path in reverse direction too
- [ ] Ice Shrine rails should break an ice block
- [ ] Fix rails sometimes sending you off in a straight line indefinitely
- [x] Player respawn points
- [x] Fix some effects playing twice
- [x] Fix enemies "falling over" all the time (#36)
- [x] Cave altars should drop imagination when interacted with
- [ ] Sensei Wu daily mission [chest](https://lu.lcdruniverse.org/explorer/objects/16552)


Uchu.World/Systems/Missions/UseSkillTask.cs
```diff
@@ -12,17 +12,23 @@ public UseSkillTask(MissionInstance mission, int taskId, int missionTaskIndex)
...
+        public async Task ReportProgress(int skillId, Lot? target = null)
         {
             if (!Parameters.Contains(skillId))
                 return;
 
+            // TODO: Check if target is correct, if applicable
+            // This is only used for the Spinjitzu Initiate achievements (1139 and 1140).
+            // They're supposed to progress when you smash one of the targets using the specified skill.
+            // if (Targets.Any(x => x != 0) && (target == null || !Targets.Contains((int) target)))
+            //     return;
```
TheNexusAvenger, line 26: Is there a reason why this is commented out?

Uchu.World/Objects/Components/ReplicaComponents/RailActivatorComponent.cs
```diff
@@ -0,0 +1,245 @@
...
+
+                var railComponent = (await ClientCache.GetTableAsync<Core.Client.RailActivatorComponent>())
+                    .First(c => c.Id == RailActivatorComponentId);
+
+                ActivatorDamageImmune = (bool) GameObject.Settings["rail_activator_damage_immune"];
+                LoopSound = (string) GameObject.Settings["rail_loop_sound"];
+                NoAggro = (bool) GameObject.Settings["rail_no_aggro"];
+                NotifyActivatorArrived = (bool) GameObject.Settings["rail_notify_activator_arrived"];
+                Path = (string) GameObject.Settings["rail_path"];
+                // TODO: investigate why PathDirection 0 makes the movement a straight line instead of the reverse path
```
TheNexusAvenger, line 71: How much was this looked into before? Anything note-worthy for anyone who looks at this later?

Uchu.StandardScripts/NinjagoMonastery/SkeletonTowers.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Numerics;
+using System.Threading.Tasks;
+using Uchu.Core;
+using Uchu.World;
+using Uchu.World.Scripting.Native;
+
+namespace Uchu.StandardScripts.NinjagoMonastery
+{
+    [ZoneSpecific(2000)]
+    public class SkeletonTowers : NativeScript
```
TheNexusAvenger: Can this be an `ObjectScript` instead of a `NativeScript`?

Uchu.StandardScripts/NinjagoMonastery/GoToNpcBodge.cs
```diff
@@ -0,0 +1,51 @@
+using System.Linq;
+using System.Threading.Tasks;
+using Uchu.World;
+using Uchu.World.Scripting.Native;
+
+namespace Uchu.StandardScripts.NinjagoMonastery
+{
+    [ZoneSpecific(2000)]
+    public class GoToNpcBodge : NativeScript
```
TheNexusAvenger, line 10: Changing this to a migrate for the CDClient was discussed before after this script was created. Was this looked into?

TheNexusAvenger (commented):
I'll start playing this locally to test out the changes. Got a few questions in the meantime.

Uchu.World/Systems/Missions/UseSkillTask.cs
enteryournamehere: I've left it commented out for now because there isn't a mechanism in place to actually pass the target to this function.
The UseSkillTask function gets called as soon as a skill is started, not when any behaviour inside the skill tree successfully smashes an object, and I couldn't find a clean way to implement this right now.
Having it commented out means players will complete achievements 1139 and 1140 by just using the skill the required amount of times (regardless of whether or not they kill Crux Prime enemies with it); uncommenting it will make the achievements impossible to complete.

Uchu.StandardScripts/NinjagoMonastery/GoToNpcBodge.cs
enteryournamehere: It was not, no. It might also be possible to get rid of this script by changing how GoToNpc tasks are handled (see also the conversation on Discord about this), but this will need to be investigated more (by eg. seeing what happens when the tasks are always regarded as complete, or if the value of target_objectID could be used)

Uchu.World/Objects/Components/ReplicaComponents/RailActivatorComponent.cs
enteryournamehere: Been a couple months since I wrote that, IIRC I ended up sending and receiving the exact same packets as I found in the captures from live, but it just didn't work. However I haven't added a handler for `RequestRailActivatorState`, maybe that's related.

Uchu.StandardScripts/NinjagoMonastery/GoToNpcBodge.cs
TheNexusAvenger: Probably should wait to get some type of automation testing going on this or for this to be a problem. Like other changes to the mission system, I changing the behavior leading to unrelated problems elsewhere.

Uchu.StandardScripts/NinjagoMonastery/GarmadonCelebration.cs
```diff
@@ -0,0 +1,33 @@
...
+            var physics = gameObject.GetComponent<PhysicsComponent>();
+
+            // Listen to player entering the trigger area
+            Listen(physics.OnEnter, (other) =>
+            {
+                if (!(other.GameObject is Player player)) return;
+
+                // Check if player has seen celebration before
+                var character = player.GetComponent<CharacterComponent>();
+                if (character.GetFlag(125)) return;
```
MickVermeulen: Wasn't there flag enum? If so, might be nice to add this for readability

Uchu.StandardScripts/NinjagoMonastery/GoToNpcBodge.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        public override Task LoadAsync()
+        {
+            // Okay look this is a really stupid script but the cdclient is messed up.
+            // Check the lu-explorer links I left here. The GoToNpc tasks have
+            // entirely unrelated targets. (These targets aren't used for any other
+            // GoToNpc tasks, so it should be fine to just call GoToNpc for them.)
+
+            // https://lu.lcdruniverse.org/explorer/missions/1797
+            const int senseiWuLot = 13789;
```
MickVermeulen: I know for sure that there's a LOT enum, for readability might be nice to store these there :)

Uchu.StandardScripts/NinjagoMonastery/LanternsOfSpinjitzu.cs
```diff
@@ -0,0 +1,29 @@
...
+            Listen(gameObject.OnInteract, async player =>
+            {
+                if (!player.TryGetComponent<MissionInventoryComponent>(out var missionInventory))
+                    return;
+                if (!player.TryGetComponent<CharacterComponent>(out var characterComponent))
+                    return;
+
+                // If player has Cole's mission active, set flag to proceed with next mission
+                if (missionInventory.HasActive(2072))
+                    await characterComponent.SetFlagAsync(2020, true);
```
MickVermeulen: Regarding to my previous comment with the flags. Ignore if there's no enum / you don't really mind

Uchu.StandardScripts/NinjagoMonastery/SkeletonTowers.cs
MickVermeulen: In theory all scripts could be object scripts

Uchu.StandardScripts/NinjagoMonastery/SkeletonTowers.cs
```diff
@@ -0,0 +1,125 @@
...
+                });
+            }
+
+            return Task.CompletedTask;
+        }
+
+        private void SetUpLegs(GameObject towerSpawnerObject)
+        {
+            var towerSpawnerComponent = towerSpawnerObject.GetComponent<SpawnerComponent>();
+            var legSpawnerComponents = new List<SpawnerComponent>();
```
MickVermeulen: Storing the game objects and then getting their components is a neater solution imo

Uchu.World/Objects/Components/ReplicaComponents/BuffComponent.cs
```diff
@@ -0,0 +1,107 @@
...
+                    RemoveBuff(buffInfo);
+                });
+            }
+
+            if (buffInfo.CancelOnDamaged)
+            {
+                // this looks ugly, maybe there's a better way to listen once
+                Delegate cancelOnDamaged1 = null;
+                Delegate cancelOnDamaged2 = null;
+                cancelOnDamaged1 = Listen(destroyable.OnArmorChanged, (newValue, delta) =>
```
MickVermeulen, line 54: Can't really think of anything than changing the api of Listen to include a `once = false` parameter

Uchu.World/Objects/GameObjects/Player.cs
```diff
@@ -57,6 +57,8 @@ internal Player()
             OnNotifyTamingBuildSuccessMessage = new Event<NotifyTamingBuildSuccessMessage>();
             OnLootPickup = new Event<Lot>();
             OnWorldLoad = new Event();
+            OnRailMovementReady = new Event();
```
MickVermeulen: Coulnd't these be moved to the game object?

MickVermeulen (commented):
There's still a few leftover logger statements, overall nice work :)

Uchu.World/Objects/Components/ReplicaComponents/BuffComponent.cs
TheNexusAvenger: We may need to do a bit more for this, like create a listener group that allows storing a set of listeners that can be disconnected at once. For this case, we listen to 2 listeners and disconnect both when received.

Example:
```C#
var listenerGroup = new ListenerGroup(maxCalls = 1); // MaxCalls may make it more clear this event can only be called once.
listenerGroup.Listen(event1, (...) => {
	...
});
listenerGroup.Listen(event2, (...) => {
	...
});

```

Uchu.StandardScripts/NinjagoMonastery/LanternsOfSpinjitzu.cs
TheNexusAvenger: It would improve readability to create an enum if we don't. There is a difference between "flag 2072" and "flag SomeMissionName".

enteryournamehere: There is a FactionFlag enum for the 4 factions, I'll change this to not be only for factions

Uchu.StandardScripts/NinjagoMonastery/SkeletonTowers.cs
enteryournamehere: Made this one an object script, there are a few non-object scripts left (e.g. LearnSpinjitzu and SkeletonBuffs) but I wasn't able to find a good way to convert those (the implementation in live was possibly not using a script)

Uchu.StandardScripts/NinjagoMonastery/GoToNpcBodge.cs
enteryournamehere: Done, thanks!

Uchu.World/Objects/GameObjects/Player.cs
enteryournamehere: Do you mean moving the event to the rail activator object, or having it still fire on the player but be defined in `GameObject.cs` instead of `Player.cs`?

PR #243, opened 2021-06-26
Add Object Scripts

TheNexusAvenger:
This pull request migrates a lot of the existing native scripts from zone-based scripts to object-based scripts (scripts that are created and cleared with objects). The following changes have also been made:
* Changed buttons to be able to be held down by players with the countdown going when the button is destroyed (quickbuild resets) or all players on the button leave. `OnInteract` is also internally handled for the buttons.
* Added script for the Nimbus Rocket creation.
* Added Avant Gardens jet effect script.
  * The airstrike skill currently doesn't work. The area of effect component of the skill ends up applying to nothing because there is no `BaseCombatAiComponent`.
* Added Return to the Venture Explorer computer mission script. This completes all the missions in Return to the Venture Explorer.
* (Potentially; didn't test before) Fixed vault ATMs not working in Nimbus Station and probably properties.
* Fixed removed character flags not saving to the database.

There was also a new `AddNewComponent` method added to allow multiple of the same type of components. This was done initially for multiple physics components but was removed after some internal discussion. The code changes end up being a minor refactor with no change.


Uchu.World/Scripting/Native/ObjectScript.cs
```diff
@@ -0,0 +1,485 @@
...
+        /// Adds a cylinder physics body to the object.
+        /// </summary>
+        /// <param name="radius"></param>
+        /// <param name="name"></param>
+        public void SetProximityRadius(float radius, string name)
+        {
+            // Add the cylinder physics body.
+            // Even if the physics is overriden, the physics body being tied
+            // to a physics component allows it to play nicely in the existing system.
+            var physics = this.GameObject.AddComponent<PhysicsComponent>();
```
enteryournamehere: Instead of overriding the physics and leaving a 'detached' physics body, you can make a new gameobject and attach the physics to that like so:
```cs
var proximityObject = World.GameObject.Instantiate(GameObject.Zone);
var physics = proximityObject.AddComponent<PhysicsComponent>();
```

Uchu.World/Scripting/Native/ObjectScript.cs
```diff
@@ -0,0 +1,485 @@
...
+        /// <param name="type">Type of the LDF entry to send.</param>
+        public void SetNetworkVar(string name, object value, byte type)
+        {
+            // Set the variable.
+            this.SetVar(name, value);
+            
+            // Broadcast the variable change.
+            Task.Run(() =>
+            {
+                Console.WriteLine("SET: " + new LegoDataDictionary()
```
enteryournamehere: This `Console.WriteLine` could be changed to `Logger.Debug` for consistency, or removed if it isn't necessary/useful anymore to keep it in

enteryournamehere (changes requested):

Uchu.World/Scripting/Native/ObjectScript.cs
TheNexusAvenger: This line was accidentally committed when I was testing. In general, if there is a `Console.WriteLine`, it is something I accidentally committed instead of me not using `Logger.Debug`.

Uchu.World/Scripting/Native/ObjectScript.cs
TheNexusAvenger: Thanks for providing the example of what to do. The requested change has been made and the script that uses this (the bus door script) still works as intended.

enteryournamehere (approved):

PR #244, opened 2021-07-07
Refactor Client Cache

TheNexusAvenger:
This pull request attempts to improve how caching of the client database is handled. Rather than running direct queries, most of which are just checking an id, a query is done to the cache for a specific id. There are 3 types of caching methods:
- Entry: The database is queried for every request for an id that isn't cached.
- Burst: The entire database is loaded into memory for up to a few seconds after the last call to it.
- Persistent: The entire database is loaded into memory permanently. This is currently only done for behaviors because of how long it takes to run queries, which results in the first of a behavior type being either slow or not work at all.

When an entire table is loaded into memory, it is first converted to a dictionary by the specified index. This is done to make finding entries with a given id a dictionary read instead of iterating over a list. From my testing:
- Starting Avant Gardens with my larger character (completed missions up to Nexus Tower with full Rank 2 armor) went from **14.867 seconds to 9.513 seconds**, measured from the first message logged to "<Player> loaded...".
- Starting Avant Gardens with my smaller character (completed missions up to Block Yard) went from **13.487 seconds to 9.534 seconds**, measured from the first message logged to "<Player> loaded...".
- Performing skills for the first time, such as equipping items, smashing Venture Explorer crates, and scripts like the monument lasers, are *much* faster.
- Performing queries is simpler to write now since it is just "find the type that matches the id" instead of "perform this LINQ statement that returns if the ids match".

Memory usage is a bit mixed from my testing. The average memory usage is significantly higher on world load, but the peak memory usage is significantly lower. Idle memory usage was about 240MB vs 310MB before and after the cache. However, I noticed peak memory usage going all the way up to 370MB without the caching changes, and never going above about 320MB with the cache changes.


Uchu.Core/Client/Attribute/CacheMethodAttribute.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// </summary>
+        public readonly CacheMethod Method;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="method"></param>
+        public CacheMethodAttribute(CacheMethod method)
+        {
+            this.Method = method;
```
MickVermeulen, line 40: you don't have to use `this` here

Uchu.World/Client/CdClient/CacheTable/BaseTableCache.cs
```diff
@@ -0,0 +1,108 @@
...
+        internal readonly PropertyInfo _cdClientContextProperty;
+
+        /// <summary>
+        /// Creates the base table cache.
+        /// </summary>
+        /// <param name="type">Type of the table.</param>
+        /// <param name="index">Index of the table.</param>
+        public BaseTableCache(Type type, PropertyInfo index)
+        {
+            this.TableType = type;
```
MickVermeulen, line 34: no need for `this
`

Uchu.World/Client/CdClient/CacheTable/EntryCacheTable.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <returns>All of the types that match the index.</returns>
+        public override async Task<T[]> FindAllAsync<T>(object index) where T : class
+        {
+            index = ConvertKey(index);
+            
+            // Populate the cache entry.
+            await this._semaphore.WaitAsync();
+            if (!this._cachedTable.ContainsKey(index))
+            {
+                // Due to Reflection, LINQ isn't suitable in this case.
```
MickVermeulen, line 59: thanks for the preliminary comment :P

MickVermeulen (commented):
overall nice refactor, I really like the new API (especially being able to just call `Find`), just had some styling comments :)

Uchu.Core/Client/Attribute/CacheMethodAttribute.cs
TheNexusAvenger: The `this` is redundant, but is it worth removing? There are a lot of other cases in the code where we use `this` when it isn't needed.

MickVermeulen: Afaik the stylegyuide used by Uchu only uses `this` when an object passes a ref to itself

TheNexusAvenger: I am not sure where we keep our style guide, or if we really should care about usage of `this`. I find it a bit more clear that it is referencing a property of an object rather than a static or local variable. And if we do care... then I have a lot of `this` to remove.

MickVermeulen: We don't have our own style guide but use the one from MS, feel free to merge if you don't really mind but it's the style guide we generally follow.

TheNexusAvenger: zaop and I may consider a reformat later for some of Wincent's old code that has a new line between most lines. `enhancement/struct-packets` probably should be merged first to automatically resolve a bunch of those. We need to figure out long-term if we want to keep using `this.` for clarity or remove it for less text.

Issue #201, opened 2021-02-16
Turrets Fire At Non-Maelstrom


TheNexusAvenger:
https://user-images.githubusercontent.com/13441476/108011952-2cc93380-6fd6-11eb-8c69-d7cbe76b18ba.mp4

Turrets, including Assembly and Avant Gardens Mechs, currently fire at non-Maelstrom, such as pet treasure and smashables. This may also be connected to the problem of skeletons at the Ninjago Monastery destroying smashables as well.

Non-mech turrets also don't register kills for the player, which may be worth a different GitHub issue.

PR #245, opened 2021-07-15
Switch SpawnQuickbuild to AuthoredGameObject

Frozenreflex:
Fixes enemies not giving drops from sentries, which is part of #201


enteryournamehere:
Good find, thank you for the PR!

PR #257, opened 2021-07-20
Improve the behavior of death planes

### LINKED ISSUE #74: Portabello's Waters
Yea if you go into the waters in this zone, you will get frozen.  Only way to get out is through punches or via /fly on command.

### LINKED ISSUE #190: Kill planes

Frozenreflex:
Fixes #190 and #74


Uchu.StandardScripts/General/DeathPlane.cs
```diff
@@ -0,0 +1,35 @@
...
+    [ScriptName("l_act_player_death_trigger.lua")]
+    public class DeathPlane : ObjectScript
+    {
+        /// <summary>
+        /// Creates the object script.
+        /// </summary>
+        /// <param name="gameObject">Game object to control with the script.</param>
+        public DeathPlane(GameObject gameObject) : base(gameObject)
+        {
+            //Console.WriteLine("Death barrier positioned at " + gameObject.Transform.Position + ", Scale of GameObject is " + gameObject.Transform.Scale);
```
TheNexusAvenger: This probably should be removed. "Permanent" messages should use `Logger.Debug` or `Logger.Information` over `Console.WriteLine` for performance and thread safety reasons.

Uchu.StandardScripts/General/DeathPlane.cs
```diff
@@ -0,0 +1,35 @@
...
+
+            var physics = gameObject.GetComponent<PhysicsComponent>();
+            if (physics == null || physics == default) return;
+            Listen(physics.OnEnter, other =>
+            {
+                if (!(other.GameObject is Player player)) return;
+                player.GetComponent<DestroyableComponent>().Health = 0;
```
TheNexusAvenger: Should `SmashAsync` be used instead of setting the health to 0?

Uchu.StandardScripts/General/DeathPlane.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// Creates the object script.
+        /// </summary>
+        /// <param name="gameObject">Game object to control with the script.</param>
+        public DeathPlane(GameObject gameObject) : base(gameObject)
+        {
+            //Console.WriteLine("Death barrier positioned at " + gameObject.Transform.Position + ", Scale of GameObject is " + gameObject.Transform.Scale);
+
+            var physics = gameObject.GetComponent<PhysicsComponent>();
+            if (physics == null || physics == default) return;
```
TheNexusAvenger: This check is the same. `default` is used to check if a value type is default in memory (`0`, `false`, etc) or is null, or just if an object type is null. Since the component is an object, this becomes `if (physics == null || physics == null) return;`.

TheNexusAvenger (commented):
Glad to see this is finally a thing considering how long many people have looked into this. Got a few comments.

Uchu.StandardScripts/General/DeathPlane.cs
Frozenreflex: I plan on removing this before it gets pushed to dev, I left it there in case I needed to use it again.

Uchu.StandardScripts/General/DeathPlane.cs
Frozenreflex: Probably, I will change it to that.

Uchu.StandardScripts/General/DeathPlane.cs
Frozenreflex: Will fix.

TheNexusAvenger:
Played around with this for a bit. For the non-properties, this is what I noticed:
* World 1000 (Venture Explorer) and part of World 1300 (Gnarled Forest): Works completely as expected.
~~* Worlds 1100 (Avant Gardens), 1200 (Nimbus Station), 1604 (Portabello), and part of World 1300 (Gnarled Forest): Missing the script `scripts\ai\AG\L_ACT_SHARK_PLAYER_DEATH_TRIGGER.lua`. May be similar to the other script except for the achivements.~~ (Edit: Didn't have that commit locally for some reason.)
* Worlds 1400 (Forbidden Valley) and 1800 (Crux Prime): Couldn't find a script, but [Object 5652](https://lu-explorer.web.app/objects/5652) appears relevant.
* World 2000 (Ninjago): Couldn't find a script, but [Object 14510](https://lu-explorer.web.app/objects/14510) appears relevant.

Frozenreflex:
>     * Worlds 1100 (Avant Gardens), 1200 (Nimbus Station), 1604 (Portabello), and part of World 1300 (Gnarled Forest): Missing the script `scripts\ai\AG\L_ACT_SHARK_PLAYER_DEATH_TRIGGER.lua`. May be similar to the other script except for the achivements.

The directory in the comment in the code is wrong, it's actually the one you mentioned, but that script should still run normally since it only cares about the name of the lua file... does it not run at all?

TheNexusAvenger:
> The directory in the comment in the code is wrong, it's actually the one you mentioned, but that script should still run normally since it only cares about the name of the lua file... does it not run at all?

Uh, oops. I didn't have that commit when testing for some reason. That can be ignored.

TheNexusAvenger:
> Worlds 1400 (Forbidden Valley) and 1800 (Crux Prime): Couldn't find a script, but Object 5652 appears relevant.
World 2000 (Ninjago): Couldn't find a script, but Object 14510 appears relevant.

Did some more digging around. We currently check `custom_script_server` in `LuaScriptComponent`, but we don't check `custom_script_client`. If we add reading this and add the relevant script(s), this will either finish death planes for all non-properties, or everything if the teleport script is added for properties.

TheNexusAvenger:
Additional changes that were made:
- Tweaked object script loading to reduce duplicate code. Check for `custom_script_server` for the falling player scripts.
- Add death planes for the remaining worlds.
- Add teleport scripts for properties.
- Fix player physics being too low, resulting in being killed on land in Portabello (1604).

Open problem:
- The falling death plane is hard-coded to use 2 seconds as a delay before smashing. May have missed something, but I couldn't figure out a better option. 2 seconds doesn't work as expected in all cases, mainly in some cases of Crux Prime (too long) and Ninjago (too short).
- The main Assembly teleporters in Nexus Tower no longer work because the physics object is underground.
  - Edit: Hacky workaround - In `PrimitiveModelComponent`, make the "Cylinder" (currently a sphere) of radius `Math.Max(cylinderSize.X, cylinderSize.Y)` instead of just `cylinderSize.Y`, which may be the smaller of the dimensions. Supporting proper Cylinder collisions is the better move though. No guarantee I will look at it.

TheNexusAvenger:
Tested the functionality of this. There are 2 outstanding issues currently:
- The shark attack in Nimbus Station no longer works.
- The fall death in Crux Prime no longer works.

The death planes in all other worlds seem to work, and the main Assembly teleporters work again.

Uchu.StandardScripts/AvantGardens/MaelstromSample.cs
```diff
@@ -1,24 +1,43 @@
...
 namespace Uchu.StandardScripts.AvantGardens
 {
     /// <summary>
-    /// Native implementation of scripts/02_client/map/ag/l_ag_maelstrom_sample.lua
+    /// Script to show/hide maelstrom samples based on whether the player has the relevant mission
     /// </summary>
-    [ScriptName("l_ag_maelstrom_sample.lua")]
-    public class MaelstromSample : ObjectScript
+    [ZoneSpecific(1100)]
+    public class MaelstromSample : NativeScript
```
TheNexusAvenger: What was the purpose of making this `ZoneSpecific` again?

Uchu.StandardScripts/AvantGardens/TargetFilter.cs
```diff
@@ -1,23 +1,42 @@
...
 
 namespace Uchu.StandardScripts.AvantGardens
 {
     /// <summary>
-    /// Native implementation of scripts/02_client/map/ag/l_ag_plunger_target.lua
+    /// Script to show/hide targets based on whether the player has the relevant mission
     /// </summary>
-    [ScriptName("l_ag_plunger_target.lua")]
-    public class TargetFilter : ObjectScript
+    [ZoneSpecific(1100)]
```
TheNexusAvenger: What was the purpose of making this `ZoneSpecific` again?

TheNexusAvenger (commented):
Changes for the death plane code looks good. Got 2 questions about 2 other changes.

Uchu.StandardScripts/AvantGardens/TargetFilter.cs
enteryournamehere: The ScriptName it used refers to a client script. After the changes on this branch, these don't get loaded anymore. No server script name is attached to these objects, so I've reverted this script to an older NativeScript implementation.

Uchu.StandardScripts/AvantGardens/MaelstromSample.cs
enteryournamehere: The ScriptName it used refers to a client script. After the changes on this branch, these don't get loaded anymore. No server script name is attached to these objects, so I've reverted this script to an older NativeScript implementation.

TheNexusAvenger: (Just going to use this now since there is 2 of the same comment)
Any thoughts about an `ObjectSpecific` or `LotSpecific` attribute so we don't have extra, duplicate event handlers?

enteryournamehere (approved):

PR #270, opened 2021-07-22
Spawn Location Persistence

### LINKED ISSUE #84: player is always rendered upside down on world load
When joining a world, during the intro screen, the player is rendered upside down.

TheNexusAvenger:
Closes #84

This pull request:
- Adds persisting spawn locations between sessions for the same world.
- Adds non-default spawn locations for landing by rockets.
- Adds item checking for alternative landing locations (such as a ninja hood to land at the tree instead of the gate).
- Fixes players not appearing in the landing animation from the Venture Explore to Avant Gardens.
- Fixes players spawning upside down on world load with default spawn locations.

The `enhancement/ag-survival` depends on this branch for setting the respawn location when leaving.

### DIFF
--- a/Uchu.World/Objects/Components/Server/RocketLaunchpadComponent.cs
+++ b/Uchu.World/Objects/Components/Server/RocketLaunchpadComponent.cs
@@ -1,35 +1,70 @@
 using System.Linq;
-using System.Threading.Tasks;
-using Microsoft.EntityFrameworkCore;
 using Uchu.Core;
+using Uchu.Core.Client;
+using Uchu.World.Client;
 
 namespace Uchu.World
 {
     [ServerComponent(Id = ComponentId.RocketLaunchComponent)]
     public class RocketLaunchpadComponent : Component
     {
+        /// <summary>
+        /// Target scene (spawn location) for landing.
+        /// </summary>
+        public string TargetScene { get; set; }
+        
+        /// <summary>
+        /// Precondition required for landing at the
+        /// alternative scene.
+        /// </summary>
+        public string AlternativeTargetScenePrecondition { get; set; }
+        
+        /// <summary>
+        /// Target scene (spawn location) for landing if the
+        /// alternate landing precondition is met.
+        /// </summary>
+        public string AlternativeTargetScene { get; set; }
+        
+        /// <summary>
+        /// Creates the component.
+        /// </summary>
         protected RocketLaunchpadComponent()
         {
             Listen(OnStart, () =>
             {
+                // Load the laucnhpad information.
+                var componentId = GameObject.Lot.GetComponentId(ComponentId.RocketLaunchComponent);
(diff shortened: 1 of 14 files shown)


Uchu.World/Objects/Zone.cs
```diff
@@ -130,7 +130,7 @@ public async Task InitializeAsync()
             Checksum = ZoneInfo.LuzFile.GenerateChecksum(ZoneInfo.LvlFiles);
             ZoneId = (ZoneId) ZoneInfo.LuzFile.WorldId;
             SpawnPosition = ZoneInfo.LuzFile.SpawnPoint;
-            SpawnRotation = ZoneInfo.LuzFile.SpawnRotation;
+            SpawnRotation = new Quaternion(ZoneInfo.LuzFile.SpawnRotation.W, ZoneInfo.LuzFile.SpawnRotation.X, ZoneInfo.LuzFile.SpawnRotation.Y, ZoneInfo.LuzFile.SpawnRotation.Z);
```
enteryournamehere: As the quaternion in LuzFile.SpawnRotation is wrong, this is something that should be fixed in InfectedRose instead of here. Additionally, the correct way to interpret it is actually `Quaternion(Y, Z, W, X)` instead of `Quaternion(W, X, Y, Z)`; it's written in the file as W, X, Y, Z, so for X the second value should be used, for Y the third value, for Z the fourth value and for W the first value.
Supported by some screenshots:
Crux Prime with Quaternion(W, X, Y, Z)
Crux Prime with Quaternion(W, X, Y, Z)
Crux Prime with Quaternion(Y, W, Z, X)
Crux Prime with Quaternion(Y, W, Z, X)

Uchu.World/Objects/Components/Server/RocketLaunchpadComponent.cs
```diff
@@ -1,35 +1,74 @@
...
+        public string AlternativeTargetScene { get; set; }
+        
+        /// <summary>
+        /// Creates the component.
+        /// </summary>
         protected RocketLaunchpadComponent()
         {
             Listen(OnStart, () =>
             {
+                // Load the laucnhpad information.
```
enteryournamehere: typo in "launchpad"

enteryournamehere (changes requested):
Very nice to see this! Only issue I have is that the correction of LuzFile.SpawnRotation shouldn't be done by Uchu but by InfectedRose, at the point it's actually reading the values from the .luz file.

Uchu.World/Objects/Components/Server/RocketLaunchpadComponent.cs
TheNexusAvenger: Nice find.

Uchu.World/Objects/Zone.cs
TheNexusAvenger: Makes sense and thanks for correcting out how the rotation should be corrected.

enteryournamehere (approved):

PR #276, opened 2021-07-27
Nested Struct Packets

TheNexusAvenger:
This pull request adds support for structs within network packets. No additional functionality is implemented yet and this was created to satisfy future uses.


enteryournamehere (approved):
Thanks for adding this - I have not yet been able to use it to implement something new (e.g. buff component serialization), but the code and tests all look good to me.

PR #277, opened 2021-07-27
Avant Gardens Survival

TheNexusAvenger:
This pull request adds Avant Gardens Survival, except:
- Global leaderboards - Will be implemented later.
- Teams - Major feature that is worth a separate pull request.
- AI movement - Major feature that is worth a separate pull request.
- Slow down spawners (enemies spawn *really* fast right now) - Haven't looked into but not specific to Avant Gardens Survival.


Uchu.StandardScripts/Base/GenericActivityManager.cs
```diff
@@ -0,0 +1,381 @@
...
+
+        /// <summary>
+        /// Sends the RequestActivitySummaryLeaderboardDataMessage message to a client.
+        /// </summary>
+        /// <param name="player">Player to send.</param>
+        /// <param name="activityId">Activity id to send.</param>
+        /// <param name="numberOfResults">Number of results to send.</param>
+        public void GetLeaderboardData(Player player, int activityId, int numberOfResults)
+        {
+            player.Message(new RequestActivitySummaryLeaderboardDataMessage()
```
enteryournamehere: Isn't this message sent from the client to the server?

Uchu.World/Scripting/Native/ObjectScript.cs
```diff
@@ -452,13 +541,98 @@ private void SendEffectsToPlayer(Player player)
...
+            {
+                Associate = this.GameObject,
+            });
+        }
+        
+        /// <summary>
+        /// Adds a player to the current minigame.
+        /// </summary>
+        /// <param name="player">Player to add.</param>
+        public void MiniGameAddPlayer(Player player)
```
enteryournamehere, line 619: MiniGameAddPlayer and AddActivityUser do the same thing, is this because there are just 2 functions for it in the lua scripts or should they be used in different circumstances? (Asking so I can use the right one in the Sentinel Combat Challenge script.)

enteryournamehere (approved):
Love it!!
One thing I found during gameplay is that quickbuilds spawned by the player don't disappear when a new round is started after the player dies. So you can spawn a turret quickbuild, die, and it'll still be there to build. Not very noticeable or important, though.

Also: the maelstrom mechs seem to not take range into account and will snipe you from across the battlefield. I'm not sure if this is a bug or if it only seems weird because they don't move like they're supposed to.

Lastly, the right part at the scoreboard UI at the top is supposed to show the next goal the player can reach, I think that's either the next achievement time or their PB, whichever is closest. Have you looked into this, does it depend on something else or will it just not be added in this PR?

TheNexusAvenger:
> One thing I found during gameplay is that quickbuilds spawned by the player don't disappear when a new round is started after the player dies. So you can spawn a turret quickbuild, die, and it'll still be there to build. Not very noticeable or important, though.

The script doesn't touch any of these, and aren't they supposed to naturally go away?

> Also: the maelstrom mechs seem to not take range into account and will snipe you from across the battlefield. I'm not sure if this is a bug or if it only seems weird because they don't move like they're supposed to.

I dumped the LDFs of all the objects at one point and saw their range is 180. Wouldn't surprise me if the value is incorrect like the factions are, but I don't know what to correct it to if this is the case. Doesn't help that so many spawn compared to what should.

> Lastly, the right part at the scoreboard UI at the top is supposed to show the next goal the player can reach, I think that's either the next achievement time or their PB, whichever is closest. Have you looked into this, does it depend on something else or will it just not be added in this PR?

Don't know for this. Nothing showed up in the packet captures that I noticed, but I may have missed something.
Edit: May depend on global leaderboards. No clue how the client would know it's personal best.

Uchu.World/Scripting/Native/ObjectScript.cs
TheNexusAvenger: They appear to be the same. Something may be missing from the implementation since the Lua script calls both of those methods in different places.

Uchu.StandardScripts/Base/GenericActivityManager.cs
TheNexusAvenger: From what I remember, when the client sees, this game message, it echoes the request back to the server. Mick says the request is used, but since we don't send any data back currently as part of global leaderboards, it doesn't do anything.

enteryournamehere:
> The script doesn't touch any of these, and aren't they supposed to naturally go away?

They do naturally go away, but you can build them before that happens. But if it isn't mentioned in the lua script then yeah they should probably just stay there.

> I dumped the LDFs of all the objects at one point and saw their range is 180. Wouldn't surprise me if the value is incorrect like the factions are, but I don't know what to correct it to if this is the case. Doesn't help that so many spawn compared to what should.

Makes sense. They seem to have quite a big range in [this footage](https://www.youtube.com/watch?v=v7O-HliCt6I) too. Difference might be that in that video they don't shoot through terrain.

> Don't know for this. Nothing showed up in the packet captures that I noticed, but I may have missed something.

Looking at the client script I think it's handled in the onSendActivitySummaryLeaderboardData handler. Would be something to implement with the rest of the leaderboard stuff later on.

Uchu.StandardScripts/Base/GenericActivityManager.cs
enteryournamehere: ah okay. Will look at this later when adding leaderboards.

Uchu.World/Scripting/Native/ObjectScript.cs
enteryournamehere: Fair enough, thanks for explaining

Uchu.StandardScripts/Base/SurvivalEnemyDeathScore.cs
```diff
@@ -0,0 +1,41 @@
...
+            this.SetNetworkVar("points", Points);
+            
+            // Set the factions of the enemy.
+            // Sentries don't have the correct factions by default for some reason.
+            if (!gameObject.TryGetComponent<DestroyableComponent>(out var destroyableComponent)) return;
+            destroyableComponent.Factions = new[] {4};
+            destroyableComponent.Enemies = new[] {1};
+            
+            // Listen to the enemy being smashed.
+            if (!GameObject.TryGetComponent<DestructibleComponent>(out var destructibleComponent));
```
enteryournamehere: I think there's a `return` missing before the `;`

TheNexusAvenger: Yep. Didn't manifest into a problem.

enteryournamehere (approved):

PR #279, opened 2021-07-29
Reduce Various Exceptions and Problems

### LINKED ISSUE #167: UseConsumable and Interact Mission Tasks Ignore Target Value
It is possible to complete "missions" (achievements) that require using consumables or interacting, but they do not rely on the target value set and only rely on 1 consumable/interact being used. I noticed this on Dizzy Tikis (id 599), Quit Yer Bellyachin' 1 (id 242), and Quit Yer Bellyachin (id 247), and I would expect this on about a dozen other Interact missions according to the database.

### LINKED ISSUE #263: Achievements not giving rewards
Goes for Venture League Privateer, Ninja Acolyte, Ninja Initiate, and all Achiever achievements (which don't pop up at all).

### LINKED ISSUE #265: Some missions are given again later
examples are Beck Strongheart's salute mission, and Coalessa's pet taming missions.

### LINKED ISSUE #274: Character rebuilds upside down
After death the character will rebuild upside down. I saw this happen at the Forbidden Valley great tree after I executed the /smash and after being killed by maelstrom horsemen.

### LINKED ISSUE #275: You can't receive an item reward if you have to choose one (ex. spear, sword)
I saw this happen on AG on the test server. If offered the choice of one reward from two+ choices when you hit check nothing happens. I saw this multiple times on different missions.

TheNexusAvenger:
Closes #167, #265, and #274
Might Close #263 and #275

This pull request attempts to address a couple of long-standing exceptions/problems, including:
- Exceptions if a mission task does not exist in the cache.
- Exceptions when starting projectiles if the firing object is destroyed.
- Exceptions when starting the zone and setting the visible objects for the player.
  - The implementation also reduces the constant `ToArray` calls to the object lists to just after the list changes. Memory usage on my system was down ~30MB (~10%) on Avant Gardens.
- Exceptions when the cache tables are indexed in a way that causes a race condition from multiple threads.
- Exceptions when arriving by rockets.
- Exceptions when trying emote-based missions without a target.
- Changing the bodge script for GoToNpc to a CDClient migration.
- Saving inventories that failed to load, leading to inventories getting wiped.
- Non-repeatable missions being offered again.
- Players appearing to stutter when moving around.
- Players respawning upside-down.
- `Interact` and `UseConsumable` missions being completed after 1 interaction instead of their target value.


Uchu.World/Client/CdClient/CacheTable/EntryCacheTable.cs
```diff
@@ -53,7 +53,7 @@ public EntryCacheTable(Type type, PropertyInfo index) : base(type, index)
             index = ConvertKey(index);
             
             // Populate the cache entry.
-            await this._semaphore.WaitAsync();
+            await _semaphore.WaitAsync();
```
enteryournamehere, line 56: is it intentional that the `this.` before `_semaphore` is removed?

TheNexusAvenger: `_semaphore` is now static. The `CdClientContext` was static before, which led to a concurrency exception in rare cases.

Uchu.World/Systems/Missions/UseConsumableTask.cs
```diff
@@ -12,7 +12,7 @@ public UseConsumableTask(MissionInstance mission, int taskId, int missionTaskInd
 
         public override MissionTaskType Type => MissionTaskType.UseConsumable;
 
-        public override bool Completed => Progress.Contains(Target);
+        public override bool Completed => Progress.Select(lot => (int) lot == Target).ToArray().Length >= RequiredProgress;
```
enteryournamehere: Because of the check on line 19, `Progress` can never contain anything other than `Target`, hence `=> Progress.Count >= RequiredProgress` should suffice. And that's actually the default value of `Completed` (see [MissionTaskInstance.cs:94](https://github.com/UchuServer/Uchu/blob/caa7be2799eacba313f0c30835e50adb3d5ddf05/Uchu.World/Systems/Missions/MissionTaskInstance.cs#L94)) so I think this line can be removed entirely.

Uchu.World/Systems/Missions/InteractTask.cs
```diff
@@ -12,7 +12,7 @@ public InteractTask(MissionInstance mission, int taskId, int missionTaskIndex)
 
         public override MissionTaskType Type => MissionTaskType.Interact;
 
-        public override bool Completed => Progress.Contains(Target);
+        public override bool Completed => Progress.Select(lot => (int) lot == Target).ToArray().Length >= RequiredProgress;
```
enteryournamehere: see comment on UseConsumableTask

Uchu.World/Client/CdClient/ClientCache.cs
```diff
@@ -51,6 +51,10 @@ public static class ClientCache
         /// </summary>
         public static async Task LoadAsync()
         {
+            // Update the CDClient database.
+            await using var cdContext = new CdClientContext();
+            await cdContext.EnsureUpdatedAsync();
```
enteryournamehere: This does take a second or more to run, could this cause issues if we were to add more migrations later?

Uchu.Core/Data/ThreadedList.cs
```diff
@@ -0,0 +1,207 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Uchu.Core.Data
+{
+    public class ThreadedList<T> : IList<T>, IDisposable where T : class
```
enteryournamehere: There is a warning on this line: `Rename Uchu.Core.Data.ThreadedList<T> to end in 'Collection'.`

enteryournamehere (changes requested):

Uchu.World/Client/CdClient/CacheTable/EntryCacheTable.cs
enteryournamehere: Ah, didn't notice that, makes sense :+1:

Uchu.World/Client/CdClient/ClientCache.cs
TheNexusAvenger: I haven't tested the effect on load times with this change. We could do a code-level change instead of a data-level change, but I am not sure about having hard-coded changes like that.

Uchu.Core/Data/ThreadedList.cs
TheNexusAvenger: I guess I can make this `ThreadedCollection<T>`. Curious why this is a warning in C#.

Uchu.World/Systems/Missions/UseConsumableTask.cs
TheNexusAvenger: Makes sense. I knew about the check for the `Target` matching, but kept it to sanity check. I think it is fine to remove for this and the other case.

Uchu.World/Client/CdClient/ClientCache.cs
enteryournamehere: I don't think we should do a hardcoded change no. But (as I understand it) at [WorldUchuServer.cs:87](https://github.com/UchuServer/Uchu/blob/caa7be2799eacba313f0c30835e50adb3d5ddf05/Uchu.World/WorldUchuServer.cs#L87) the table cache is loaded on a separate thread from the one that loads the zone objects, so I was wondering if this could be a problem if it's trying to use the ClientCache before the migrations have completed.

TheNexusAvenger: Fair point. There actually would be a problem if migration was required to the Zone table since the first read is always done before the CdClient cache is created. I've moved it earlier in the process so that won't happen.

enteryournamehere (approved):

PR #281, opened 2021-08-07
Stat Pickup Achievements

### LINKED ISSUE #273: Medic! Tanks very much! and Imagine that! achievements don't work
These achievements don't work at all, and the object rewards they give don't dispense the drops and load in funny.

Frozenreflex:
Fixes #273


Uchu.World/Objects/GameObjects/GameObject.cs
```diff
@@ -148,12 +148,19 @@ protected GameObject()
...
                 foreach (var (objectScriptName, objectScriptType) in Zone.ScriptManager.ObjectScriptTypes)
                 {
                     if (!scriptName.EndsWith(objectScriptName)) continue;
                     this.Zone.LoadObjectScript(this, objectScriptType);
+                //    a = true;
                     break;
                 }
+                /*
+                if (!a && scriptName != ""){
+                    Console.WriteLine("Did not load script: " + scriptName + " Object LOT: " + Lot.Id);
```
TheNexusAvenger: Was this meant to be commited. Shouldn't really have commented out logging statements like this.

Uchu.StandardScripts/General/StatPickups.cs
```diff
@@ -13,58 +14,76 @@ public override Task LoadAsync()
                 Listen(player.OnLootPickup, lot =>
                 {
                     var stats = player.GetComponent<DestroyableComponent>();
-                    
+                    var skill = player.GetComponent<SkillComponent>();
+                    var missionInventoryComponent = player.GetComponent<MissionInventoryComponent>();
+                    var skillExecute = 0;
```
TheNexusAvenger: For this, you may find a dictionary lookup (`Dictionary<Lot, int>`) is easier to understand and shorter than this switch statement.

Uchu.World/Objects/GameObjects/GameObject.cs
Frozenreflex: This was something unrelated I was working on that I accidentally left in. It could potentially be a useful logging statement, should I switch it to the standard logger method?

TheNexusAvenger: I would say it is useful as a `Logger.Debug` statement.

Uchu.StandardScripts/General/StatPickups.cs
```diff
@@ -8,63 +10,34 @@ public class StatPickups : NativeScript
...
-                            stats.Armor += 10;
-                            break;
-                        default:
-                            return Task.CompletedTask;
-                    }
-                    
+                    var skill = player.GetComponent<SkillComponent>();
+                    var missionInventoryComponent = player.GetComponent<MissionInventoryComponent>();
+                    skill.CalculateSkillAsync(pickups[lot], player);
+                    missionInventoryComponent.StatPickupsAsync(pickups[lot]);
```
TheNexusAvenger: Should we do the StatPickup mission task updates here or in the same place as part of wherever we update skill mission tasks? They are practically the same. Alternatively, `StatPickupsAsync` could be merged with `UseSkillAsync`.

Uchu.World/Objects/GameObjects/GameObject.cs
```diff
@@ -148,12 +148,19 @@ protected GameObject()
                     this.GetComponent<LuaScriptComponent>() != default) return;
                 var scriptName = ((string) scriptNameValue).ToLower();
                 Logger.Debug($"{this} -> {scriptNameValue}");
+                var a = false;
```
TheNexusAvenger: Can this variable name be changed to something more specific, like `scriptLoaded`?

Uchu.StandardScripts/General/StatPickups.cs
Frozenreflex: Given how it's triggered by a Skill anyway, I'll merge it with UseSkillAsync

Uchu.World/Objects/GameObjects/GameObject.cs
Frozenreflex: Will do.

Uchu.StandardScripts/General/StatPickups.cs
enteryournamehere: Do make sure there aren't any items or weapons that also use one of these skills to restore stats

Uchu.StandardScripts/General/StatPickups.cs
```diff
@@ -8,63 +10,33 @@ public class StatPickups : NativeScript
     {
         public override Task LoadAsync()
         {
+            var pickups = new Dictionary<int, int>()
```
enteryournamehere: These don't need to be hardcoded, they are defined in the CDClient: the objects have an attached skill with `castOnType = 2` (table ObjectSkills).
There are actually a few more objects like this with a `castOnType = 2` skill: 7680, 8200 and 8208 (powerups for racing).
(We have a `SkillCastType` enum that should be extended to include `OnCollect` (or a similar clear name) with value `2`.)

enteryournamehere:
IMO it would be nice to use the same name that lu-explorer uses for this mission task type, which is `CollectPowerup`, to avoid any confusion later on.

Issue #289, opened 2021-08-13
Rockets stack up


teccheck:
If you build a new custom rocket it stacks on top of the old/current rocket, no matter what parts you use. I watched a [video](https://www.youtube.com/watch?v=FYqjZBnuBIg&list=PL9EFB8500E625ADC7&index=20) of the original gameplay and this is not the intended behavior. This is probably related to #63 and #157.

I don't know the code responsible for that system, but I suspect that there needs to be a check for the containing parts when putting a rocket into the inventory, so that a new instance will be created.

teccheck:
The problem is not that there is a missing check, it's that `stackSize` has the value `2147483647` when its checked in `AddLotAsync` in `InventoryManagerComponent.cs:`. So there is probably an error while getting the stack size, because it has the right value in the database

teccheck:
This code [snippet](https://github.com/UchuServer/Uchu/blob/8b579acaebcb13428b19482ce9a4c8133c34c7da/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs#L361) is the source of this problem:
```c#
// Some items have no stack size, like bricks
var stackSize = itemComponent.StackSize ?? 1;
if (stackSize == default) 
    stackSize = int.MaxValue;
```
If `stackSize` is 0 then it will be set to `int.MaxValue`. That is all good for bricks, as the comment explains but rockets do have an **actual** stackSize of 0.

I would rather check if the item is of type `Brick`, then set stackSize to `int.MaxValue`. You could also check for type `Vehicle` and set the stackSize to 0.

enteryournamehere:
I believe the issue is because Uchu doesn't have subkeys implemented, which are meant to distinguish modular build items (rockets, cars) from each other. It is used in the GameMessages for moving items in inventories (for example see [AddItemToInventoryClientSync](https://github.com/UchuServer/Uchu/blob/caa7be2799eacba313f0c30835e50adb3d5ddf05/Uchu.World/Packets/GameMessages/Server/AddItemToInventoryClientSyncMessage.cs#L20), it's also in the GMs for moving items between inventories, deleting an item from an inventory and for notifying the client about a reward that has been mailed to them).

teccheck:
I'm not quite sure... For testing I added an if statement to directly create a new stack when a rocket is added to the inventor and it worked.

Btw: Do you know if there are any item types that have a stackSize of 0 when you can have an unlimited amount of them (like bricks)?

[...]

PR #290, opened 2021-08-13
Fix stacking of rockets

teccheck:
Description: See #289

I'm open to any suggestions for this change

### DIFF
--- a/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
+++ b/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
@@ -363,8 +363,7 @@ namespace Uchu.World
             if (stackSize == default) 
                 stackSize = int.MaxValue;
             
-            // If ItemType is Vehicle
-            if(itemComponent.ItemType == 22)
+            if(settings != null)
                 stackSize = 1;
 
             var totalAdded = 0L;
--- a/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
+++ b/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
@@ -363,6 +363,10 @@ namespace Uchu.World
             if (stackSize == default) 
                 stackSize = int.MaxValue;
             
+            // If ItemType is Vehicle
+            if(itemComponent.ItemType == 22)
+                stackSize = 1;
+
             var totalAdded = 0L;
             var totalToAdd = count;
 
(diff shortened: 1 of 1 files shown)


enteryournamehere:
@MickVermeulen or @Wincent01 I would like to hear your thoughts on this, see also the comments in #289. Is this vehicle-stack-size-1 check correct, or should stacking be prevented by items having different subkeys?

MickVermeulen:
I think the best bet would be to uniquely store any item that has LDF data (which rockets and race cars do) as otherwise it's possible to overwrite this data. So I'd implement this to check whether or not there's LDF and refuse stacking if so.

teccheck:
So you mean checking if `ItemComponent.Settings` is not null in which case the method would just skip the stacking part?

MickVermeulen:
Just alter your current commit to check for the settings instead of checking for the item type. Make sure to test it, though :P

teccheck:
Should I add another commit to this pull request or should I create a new one without the first commit?

MickVermeulen:
Add another :)

teccheck:
Hi, I don't want to annoy anyone... But is there anything left to do on this pull request?

enteryournamehere:
If it's ready now, I can test it tomorrow to make sure it works well, after that's done it can be merged

teccheck:
I'd say it's ready

Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
```diff
@@ -363,6 +363,9 @@ private static async Task<ItemComponent> GetItemComponentForLotAysnc(Lot lot)
             if (stackSize == default) 
                 stackSize = int.MaxValue;
             
+            if(settings != null)
```
enteryournamehere: Small [style](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions) comment: there should be a space between `if` and the parentheses

enteryournamehere (commented):
Tested it, it works well :ok_hand:
Just one small comment, after that's updated it can be merged (unless @TheNexusAvenger has any other comments).

Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
TheNexusAvenger: The change seems small and simple, and you tested it. I don't have anything else to comment. Just that styling comment you left.

teccheck:
Ok, done :)

enteryournamehere:
Perfect thanks!

PR #292, opened 2021-08-24
Leaderboards and footraces

enteryournamehere:
- Implement leaderboards, including separate weekly/all-time sections.
  - Friend-only leaderboards are not yet implemented as Friends functionality is currently not complete.
- Add footrace script.
- Add Monument race script.
- Add a way to override objects to bypass the render distance filter; without this, the client isn't notified of a change in activity participants if they're out of range. More specifically, the client still thinks you're in a footrace if you end it while far away from the place where you start.


MickVermeulen (approved):
Nice work :D

Uchu.Core/Serializable/Struct/Property/PacketProperty.cs
```diff
@@ -31,7 +31,7 @@ public class PacketProperty : IPacketProperty
                         var data = (LegoDataDictionary) o;
                         if (data != null)
                         {
-                            var ldf = data.ToString();
+                            var ldf = data.ToString("\n");
```
TheNexusAvenger, line 34: Did this change have any effects beyond leaderboards? Pretty sure there was another script somewhere that had 2 or more LDF entries.

TheNexusAvenger (approved):
Thanks for covering this change instead of me having to do with Avant Gardens Survival. Surprised how many "LU moments" appeared with this, but somehow not at the same time.

Uchu.Core/Serializable/Struct/Property/PacketProperty.cs
enteryournamehere: The player list you see when you load into AGS with 2+ players will now also highlight your own name correctly. I don't remember if there were any more places where this made a difference (but it didn't break anything in any case).

PR #293, opened 2021-08-25
Use External Logging Library

TheNexusAvenger:
This pull request moves our currently logging code to an external library, including a few enhancements
* Add line wrapping for long messages.
* Improve the design of the logging code.
* Allow additional information to be stored with logs, like the server that generated the log entry. Mainly needed for combined logs, like when deployed in a single Docker container.

The reason this was made a "Nexus" library instead of a "Uchu" library is that I intend to use this for a few other projects, although it will be some time before they are public (trying to do a month-long stability test with the first one).


enteryournamehere (approved):
Nice changes, I especially like that it now breaks lines at spaces so full words aren't split up (where possible).

Issue #240, opened 2021-06-21
Build zone doesn't auto close on first time use


MustacheMan2:

teccheck:
Do you mean the build zone for rockets or cars? When should it auto close?

enteryournamehere:
https://user-images.githubusercontent.com/11255568/133297534-85eef468-5b3e-4508-bee1-50fc60195b82.mp4

When you click the blue Done button, you should exit the build mode, but currently this doesn't always work: in this video I still need to press Exit afterwards to actually leave the build zone. Note that this doesn't consistently happen: I just tried it on a local server and on an external one, and on the local one I _did_ automatically exit after pressing Done while on the external one I did not. It's possible that there's a race condition somewhere.

teccheck:
I see. Gonna look into it

enteryournamehere:
Forgot to add this earlier: it applies to car building too (although that's a bit broken in other ways), and actually I believe also to pet taming: see [this video](https://youtu.be/5BPo5o04-74?t=45), it's the same state that you're in when building a rocket/car. But when in pet taming mode, there's no exit button, so you're just stuck after submitting the name.

teccheck:
Do you know if this happened on Micks server? It would be cool to have logs of this happening. I haven't had this problem while I was testing my code (and I tested a lot)

enteryournamehere:
Yes it did. I can reproduce and get you some logs later today or tomorrow. Do you want to add any additional logging besides what's currently being logged on the `dev` branch (at the Debug level)?

teccheck:
Maybe you could log what functions are called in `ModularBuilderComponent`

teccheck:
I guess I already fixed this... I can reproduce this on `dev` but not on `enhancement/modular-building`

enteryournamehere:
Not sure how, but it does indeed appear to be fixed on your branch, I can't reproduce it there either :+1:

teccheck:
I guess the problem was that if you dropped a rocket onto the builder it wouldn't move the parts into `TempModels` so the `FinishBuild` method wouldn't find them. I think I fixed that in the commit `Implement StartArrangingWithModelMessage`

Issue #157, opened 2021-01-11
StartArrangingWithModelMessage is not implemented


MickVermeulen:
The StartArrangingWithModelMessage should move the provided model to the TemporaryModels inventory and disassemble it, currently it's possible to have dupe rockets as this message is ignored so the active rocket isn't removed from the player inventory.

MickVermeulen:
similar to #63

teccheck:
I tried it using the the `InventoryManagerComponent.MoveItemBetweenInventoriesAsync` Method. The problem is, if I drop a rocket on the builder, the components of that rocket fly back into my inventory. I don't know why... 

Is there a listener on the Inventory that does that?

Edit: I found the "problem". It is in `Item.SetCountAsync`. This function disassembles the item into its parts.

teccheck:
Does anyone know what happens if I press ESC, after putting a rocket into the builder (and changing it)? Right now you get the parts back into your inventory, but I guess you should get the rocket back how it was before editing it.

Issue #63, opened 2020-09-21
When placing a full rocket in a build area the rocket is not taken


Jettford:

Jettford:
and the mission that this is for doesn't complete (Mardolfs rocket mission)

Jettford:
Rockets will stack when this happens

enteryournamehere:
Fixed in #297

Issue #21, opened 2020-03-04
Rocket building can induce a client crash


elenawinters:
Rocket building can induce a client crash.

Steps to reproduce:
1. Enter building mode.
2. Drag a rocket piece out of your inventory, and place it on the ground.
3. Pick it up with shift. You will notice that nothing will appear in your hands but you will be holding something.
4. Press shift again, so you can interact with buttons.
5. Hit the Exit X that is above the rocket hologram.

The client will crash.

This wont likely be an issue for everyone, as people are likely to drag and drop rocket pieces into the hologram.

VaiskiKP:
[Video showing this happening](https://youtu.be/pmcKJ5GLu_E)

Jettford:
It might need to put the item back in your inventory not too sure, will test on another server later.

teccheck:
This should be fixed in the Modular building PR

PR #297, opened 2021-09-10
Modular Building

teccheck:
With this PR I'd like to implement a few functions from the modular building (rockets and cars). 

These are the things that I know worked in live but do not in Uchu (atm):
* [x] Dropping a model into the builder when in build mode  (#157) 
* [ ] Dropping a model into the builder when not in build mode (#63)
* [ ] Placing parts you are holding on the ground/on the rocket (#21)
* [x] Picking parts up from the ground
* [x] Show flying loot when closing the build mode
* [x] Always auto close build border after finish #240
* [x] Getting a car model if you build a car

### DIFF
--- a/Uchu.World/Handlers/GameMessages/ModularBuildingHandler.cs
+++ b/Uchu.World/Handlers/GameMessages/ModularBuildingHandler.cs
@@ -32,9 +32,9 @@ namespace Uchu.World.Handlers.GameMessages
         }
 
         [PacketHandler]
-        public void ModelPrefabHandler(StartArrangingWithModelMessage message, Player player)
+        public async Task ModelPrefabHandler(StartArrangingWithModelMessage message, Player player)
         {
-            
+            await player.GetComponent<ModularBuilderComponent>().StartBuildingWithModel(message.Item);
         }
 
         [PacketHandler]
--- a/Uchu.World/Handlers/GameMessages/ModularBuildingHandler.cs
+++ b/Uchu.World/Handlers/GameMessages/ModularBuildingHandler.cs
@@ -46,8 +46,8 @@ namespace Uchu.World.Handlers.GameMessages
         [PacketHandler]
         public void SetLastCustomBuildHandler(SetLastCustomBuildMessage message, Player player)
         {
-            if (!player.TryGetComponent<CharacterComponent>(out var characterComponent));
-            characterComponent.Rocket = message.Tokens;
+            if (!player.TryGetComponent<CharacterComponent>(out var characterComponent))
+                characterComponent.Rocket = message.Tokens;
         }
 
         [PacketHandler]
--- a/Uchu.World/Handlers/GameMessages/ModularBuildingHandler.cs
+++ b/Uchu.World/Handlers/GameMessages/ModularBuildingHandler.cs
@@ -46,7 +46,7 @@ namespace Uchu.World.Handlers.GameMessages
         [PacketHandler]
         public void SetLastCustomBuildHandler(SetLastCustomBuildMessage message, Player player)
         {
-            if (!player.TryGetComponent<CharacterComponent>(out var characterComponent))
+            if (player.TryGetComponent<CharacterComponent>(out var characterComponent))
                 characterComponent.Rocket = message.Tokens;
         }
 
--- a/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
+++ b/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
@@ -610,12 +610,14 @@ namespace Uchu.World
         /// <param name="destination">The destination inventory to move to</param>
         /// <param name="silent">Whether to send inventory update messages or not</param>
         /// <param name="showFlyingLoot">Whether to show flying loot animation</param>
-        /// <param name="objectId">The objectId that the moved item should have if it's moved as a whole stack</param>
         public async Task MoveItemBetweenInventoriesAsync(Item item, uint count, InventoryType source,
-            InventoryType destination, bool silent = false, bool showFlyingLoot = false, ObjectId objectId = default)
+            InventoryType destination, bool silent = false, bool showFlyingLoot = false)
         {
             if (item == null)
                 return;
+
+            // Use old object ID for new item if full stack (or more) is removed
+            var objectId = item.Count <= count ? item.Id : default;
 
             await RemoveItemAsync(item, count, source, silent);
             await AddLotAsync(item.Lot , count, item.Settings, destination, lootType: LootType.Inventory, showFlyingLoot: showFlyingLoot, objectId: objectId); // TODO: find out if lootType is correct (what's LootType.Relocate?)
--- a/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
+++ b/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
@@ -396,6 +396,7 @@ namespace Uchu.World
                 {
                     var toAdd = (uint) Min(stackSize, (int) totalToAdd);
                     var item = await Item.Instantiate(GameObject, lot, inventory, toAdd, extraInfo: settings, rootItem: rootItem, objectId: objectId);
+                    objectId = default;
 
                     totalToAdd -= toAdd;
                     totalAdded += toAdd;
--- a/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
+++ b/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
@@ -331,10 +331,11 @@ namespace Uchu.World
         /// <param name="rootItem">An optional parent item</param>
         /// <param name="lootType">The reason this lot is given to the player</param>
         /// <param name="showFlyingLoot">Whether to show flying loot animation</param>
+        /// <param name="objectId">The objectId that the moved item should have if it is moved as a whole stack</param>
         /// <returns>The list of items that were created while adding the lot</returns>
         public async Task AddLotAsync(Lot lot, uint count, LegoDataDictionary settings = default,
             InventoryType inventoryType = default, Item rootItem = default, LootType lootType = default,
-            bool showFlyingLoot = true)
+            bool showFlyingLoot = true, ObjectId objectId = default)
         {
             var createdItems = new List<Item>();
             
@@ -394,7 +395,7 @@ namespace Uchu.World
                 while (totalToAdd > 0)
                 {
                     var toAdd = (uint) Min(stackSize, (int) totalToAdd);
-                    var item = await Item.Instantiate(GameObject, lot, inventory, toAdd, extraInfo: settings, rootItem: rootItem);
+                    var item = await Item.Instantiate(GameObject, lot, inventory, toAdd, extraInfo: settings, rootItem: rootItem, objectId: objectId);
 
                     totalToAdd -= toAdd;
                     totalAdded += toAdd;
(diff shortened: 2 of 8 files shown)


teccheck:
Right now I'm stuck at the second point, because the client doesn't show the rocket even though it's moved into the temp-models inventory. There must be some way to tell the client but I just can't figure out how...

Jettford:
Maybe you need the ModularAssemblyComponent, I think that is a replica component that you need.

teccheck:
Do you mean ModularBuilderComponent?

enteryournamehere:
I believe Jett is referring to [this](https://lcdruniverse.org/lu_packets/lu_packets/raknet/client/replica/module_assembly/index.html) replica component (also in [lu_replica_packets docs](https://docs.google.com/document/d/1V_yhtj91QG0VBfMnmD5zC44DXwCRqjbBN98HoXXC7qs/edit#heading=h.xse7mjmay9p7#))

teccheck:
I understand this as following: some or all components on a gameobject are serialized and sent to the client? That would mean I need to implement a struct with the same properties as the live implementation. That would then fix (some of) the problems?

enteryournamehere:
> I understand this as following: some or all components on a gameobject are serialized and sent to the client? That would mean I need to implement a struct with the same properties as the live implementation.

Right. Replica components are serialized and sent to the client. Uchu has 2 ways to go about this, either by manually writing the serialization (see for example the [inventory component](https://github.com/UchuServer/Uchu/blob/caa7be2799eacba313f0c30835e50adb3d5ddf05/Uchu.World/Objects/Components/ReplicaComponents/InventoryComponent.cs#L252)), or using a struct (see for example the [bouncer component](https://github.com/UchuServer/Uchu/blob/caa7be2799eacba313f0c30835e50adb3d5ddf05/Uchu.World/Objects/Components/ReplicaComponents/BouncerComponent.cs#L29) and its [struct](https://github.com/UchuServer/Uchu/blob/caa7be2799eacba313f0c30835e50adb3d5ddf05/Uchu.World/Packets/Replica/Serialize/BouncerSerialization.cs), and the [vendor component](https://github.com/UchuServer/Uchu/blob/caa7be2799eacba313f0c30835e50adb3d5ddf05/Uchu.World/Objects/Components/ReplicaComponents/VendorComponent.cs#L44) and its [struct](https://github.com/UchuServer/Uchu/blob/caa7be2799eacba313f0c30835e50adb3d5ddf05/Uchu.World/Packets/Replica/Serialize/VendorSerialization.cs))

> That would then fix (some of) the problems?

Possibly - I don't know if this is indeed the component that should be implemented to fix this issue.
It's component no. 61 and is attached to objects 6416, 6577, 8092 and 9827. These are all modular assembly items (rocket, cars).

teccheck:
I ran into yet another problem. Its about the `Item` class. At the moment it seems that items are handled a bit differently than other GameObjects. 

For example when an item is instantiated it doesn't automatically add its components based on the`ComponentRegistry` table although I think that can be fixed quite easily. There is a method called `InitializeComponents` in `GameObject`
 that handles all of that.

As well as that items don't register themselves in the zone (like player objects do) which means that none of the `Construct` methods of the replica components get called and no SerializeMessage is sent. That is needed for the `ModularAssemblyComponent` (and possibly others) to work properly.

At least that's what I think. I'm not really familiar with the code base.

I tried calling `Construct(instance)` in `Item.Instantiate` to see if it would send a ConstructionPacket but it didn't. Should items even send those packets?

enteryournamehere:
You're right - items currently don't have components initiated. I've looked at that briefly before, adding `instance.InitializeComponents();` on [this line](https://github.com/UchuServer/Uchu/blob/dev/Uchu.World/Objects/GameObjects/Item.cs#L87) should fix it iirc.

I'm not sure about the construction of items. Maybe someone else has something useful to add about that, otherwise I can look into it later this week.

teccheck:
I've not see any Serialisation/Creation packet for a rocket in the captured packets. [The docs](https://docs.lu-dev.net/en/latest/replica.html) say that the `ModularAssemblyComponent` was sent but I'm really not sure...

teccheck:
By the way the latest commit (dbe4fa1) may not be the best solution. Feedback is welcome.

teccheck:
I'm now gonna leave out the remaining 2 Tasks because I just can't figure out what to implement to make them work. For this this reason it would be cool to get some feedback on the code from active devs if possible. I'm sure there are some things that could be improved.

Edit: I don't want to be passive aggressive, but I'm just kinda impatient ;)

Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
```diff
@@ -166,5 +182,35 @@ public async Task ConfirmFinish()
...
+                { 8044, 32 }, // "UGG - Modular Car Garage 6 (current newest)" (Cars)
+                { 9861, 32 }, // "Build Border Gnarled Forest Car" (Cars - Probably unused) 
+                { 9980, 8 },  // "Build Border for Nimbus Station" (Rockets)
+                { 10047, 8 }, // "Build border for LUPs station Rocket" (Rockets - Probably unused)
+            };
+
+            return dict[lot];
+        }
+
+        private static Lot GetModelLotForBuildMode(int mode)
```
enteryournamehere: I think it would be nice to store build modes as an enum instead of an int. You can also add the `Lot`s to `Lot.cs` as consts, so you can refer to them with, for example, `Lot.BuildBorderNimbusStation`.
Additionally I don't think a dict is the best thing to use here, a switch statement might make more sense, so you would end up with something like this:
```cs
switch(lot) {
    case Lot.NewRocketBay:
    case Lot.NimbusRocketBuildBorder:
    case Lot.LupRocketBuildBorder:
        return BuildMode.Rocket;
    case Lot.CarBuildBorder:
    case Lot.GnarledForestCarBuildBorder:
        return BuildMode.Car;
    default:
        return BuildMode.NotCustomized;
}
```
(Don't know what NotCustomized does & if that default value makes any sense - maybe Rocket or something else is better)

Uchu.World/Handlers/GameMessages/ModularBuildingHandler.cs
```diff
@@ -46,8 +46,8 @@ public async Task BuildExitConfirmationHandler(BuildExitConfirmationMessage mess
         [PacketHandler]
         public void SetLastCustomBuildHandler(SetLastCustomBuildMessage message, Player player)
         {
-            if (!player.TryGetComponent<CharacterComponent>(out var characterComponent));
-            characterComponent.Rocket = message.Tokens;
+            if (!player.TryGetComponent<CharacterComponent>(out var characterComponent))
```
enteryournamehere: The line below only runs if the player does _not_ have a Character component, I think this `!` shouldn't be here

Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
```diff
@@ -394,7 +395,7 @@ private static async Task<ItemComponent> GetItemComponentForLotAysnc(Lot lot)
                 while (totalToAdd > 0)
                 {
                     var toAdd = (uint) Min(stackSize, (int) totalToAdd);
-                    var item = await Item.Instantiate(GameObject, lot, inventory, toAdd, extraInfo: settings, rootItem: rootItem);
+                    var item = await Item.Instantiate(GameObject, lot, inventory, toAdd, extraInfo: settings, rootItem: rootItem, objectId: objectId);
```
enteryournamehere, line 398: I believe this could result in multiple stacks with the same ObjectId being created (when `count > stackSize` and `objectId` is specified), and that an `objectId = null;` line after this one should fix that

enteryournamehere (changes requested):
Looking good so far! I've left a couple comments but overall it looks great.
One thing I found when testing is that if I drop a whole rocket into the build area and then replace one of the parts with a different one from my inventory, I don't get the part that it's replacing back until I leave the build mode. Would it be possible to fix that? Replacing an individually placed part does work as expected.

teccheck:
The thing that you found is intentional. I did it that way so you wouldn't lose your rocket. But if you think the UX is bad i can change that.

Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
teccheck: The reason i did it that way is because LU did it that way. I found this [here](https://github.com/lcdr/lu-explorer/blob/1dca93fe5125aa8b2c127564e88b216a11a303ae/src/app/objects/components/item-component/item-component.component.html#L112). Still I can change it. 

Btw I too am not sure what "NotCustomized" is for. There are also some other modes

Uchu.World/Handlers/GameMessages/ModularBuildingHandler.cs
teccheck: I'll change that. Maybe someone forgot a `return;` there.

Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
teccheck: Done

Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
enteryournamehere: Both methods achieve the same goal and have the same outputs for the same inputs. The only difference is that the one using a dict throws an exception when an invalid key is supplied, while `switch` + `default` prevents that. The `switch` method is also [about 24x as fast](https://dotnetfiddle.net/0EPq7n). (Not that that matters much, though; in that example it still runs 10k times in 1.5ms)

enteryournamehere:
Hmm, makes sense. Personally I do think it would be better to always give back the part you're replacing when working on a rocket. With the current implementation the part will be inaccessible until you quit building, so you can't drag a rocket into the build area, make two changes and then revert one of those; to use that part again you'd have to exit build mode, enter it again and load the rocket again.

Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
```diff
@@ -69,78 +59,99 @@ public async Task StartBuildingAsync(StartBuildingWithItemMessage message)
...
+        /// </summary>
         public async Task Pickup(Lot lot)
         {
+            if (!(GameObject is Player player)) return;
+
             var inventory = GameObject.GetComponent<InventoryManagerComponent>();
             
-            var item = inventory[InventoryType.TemporaryModels].Items.First(i => i.Lot == lot);
+            var item = inventory.FindItem(lot, InventoryType.TemporaryModels);
+            if (item == default || item == null)
```
enteryournamehere: This could be written as just `item == default`; the `default` for `Item` [is null because it is a reference type](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/default-values)

teccheck:
Ok so then I'll change the behavior to just disassemble the rocket when the player puts it into the builder and if he quits he'll get the parts back

Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
teccheck: There is already a `BuildMode` Enum. I'm not quite sure what purpose it has. Should I add the entries to it?

Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
```diff
@@ -153,6 +163,53 @@ public async Task ConfirmFinish()
...
+            CurrentModel = null;
+
+            // Check if all parts are still available in the players inventory
+            for (var i = 0; i < parts.Length; i++)
+            {
+                Lot lot = model[i];
+
+                var item = inventory.FindItem(lot, InventoryType.TemporaryModels);
+                if (item == null)
+                    item = inventory.FindItem(lot, InventoryType.Models);
```
enteryournamehere: You can replace the if statement with just a `??`:

```suggestion
                var item = inventory.FindItem(lot, InventoryType.TemporaryModels) 
                           ?? inventory.FindItem(lot, InventoryType.Models);
```

enteryournamehere (commented):
Just one small comment but it's working perfectly :+1: love it

teccheck:
OK, done :rocket:

enteryournamehere (approved):
:rocket: indeed - :car:, even

thank you for all these improvements!

PR #298, opened 2021-09-11
Prepare For Releases

TheNexusAvenger:
This pull request attempts to add the last requirements to make releases. The changes include:
* Creating a single publish script for Windows-x64, macOS-x64, and Linux-x64. ARM64 can be added later when .NET 6 releases next year.
  * Releases will need to be made on a non-Windows system to ensure file permissions are correct.
* Checking for `Uchu.Instance.dll` and `Uchu.StandardScripts.dll` in the same directory as `Uchu.Master` for the default configuration.
* Removing the requirement for `dotnet` to be installed for launching servers.
* Disabling Redis by default since we no longer document it or recommend it.

### DIFF
--- a/Uchu.Char/Packets/Client/CharacterListRequest.cs
+++ b/Uchu.Char/Packets/Client/CharacterListRequest.cs
@@ -1,3 +1,4 @@
+using RakDotNet.IO;
 using Uchu.Core;
 
 namespace Uchu.Char
@@ -7,5 +8,10 @@ namespace Uchu.Char
         public override RemoteConnectionType RemoteConnectionType => RemoteConnectionType.Client;
 
         public override uint PacketId => 0x2;
+
+        public override void Deserialize(BitReader reader)
+        {
+
+        }
     }
 }
--- a/Uchu.Core/Cache/DatabaseCache.cs
+++ b/Uchu.Core/Cache/DatabaseCache.cs
@@ -95,7 +95,7 @@ namespace Uchu.Core
             await using var ctx = new UchuContext();
             
             string key;
-            var timeout = 1000;
+            var timeout = 5000;
             
             while (!_keys.TryGetValue(endpoint.ToString(), out key))
             {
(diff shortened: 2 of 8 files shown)


enteryournamehere:
I've updated the publish.py script to 
- directly put the output files in the `bin` directory, instead of letting dotnet put them where it wants and then trying to find the right directory (which failed when an old netcoreapp3.1 directory still existed)
- include LICENSE, README.md and Configuration.md in the archive

One thing that still bugs me is that the output contains a _lot_ of files (260 to 300, depending on the platform). It looks cluttered, it's not immediately obvious which files are relevant for the user. .NET supports making single-file executables by specifying `-p:PublishSingleFile=true`, but this results in a) no DLL being created for Uchu.StandardScripts, and b) an increased total output size (160 vs 106 MB for Linux). Alternatively: it should also be possible to ["hide" all the dependencies in a subfolder](https://stackoverflow.com/a/47867045/16759257), but I haven't gotten this to work so far.

luxaritas:
Thoughts about setting up releases to be built using GitHub Actions?

TheNexusAvenger:
> One thing that still bugs me is that the output contains a lot of files (260 to 300, depending on the platform). It looks cluttered, it's not immediately obvious which files are relevant for the user. .NET supports making single-file executables by specifying -p:PublishSingleFile=true, but this results in a) no DLL being created for Uchu.StandardScripts, and b) an increased total output size (160 vs 106 MB for Linux). Alternatively: it should also be possible to "hide" all the dependencies in a subfolder, but I haven't gotten this to work so far.

The other changes seem good, and this one I can see being worth it. Finding `config.xml` and `Uchu.Master.exe` was much more annoying than they should have been. I'll take a look at this as well. Maybe the first development release that has this will give some clues.

> Thoughts about setting up releases to be built using GitHub Actions?

We've discussed this a bit internally before but haven't decided on anything. Unlike some of the others projects I have worked on, like Nexus LU Launcher, Uchu can compile on Linux for Windows with no issues, so that isn't a limit. For now, I am going to say we do that as a separate PR in the (hopefully near) future just in case we decide not to go with it.

enteryournamehere (approved):

PR #299, opened 2021-09-17
Automatically skip Block Yard missions

enteryournamehere:
Makes the process a bit smoother for players. You'll no longer have to type /complete 6 times for a bunch of not-implemented missions. These scripts should be removed when the Spider Queen fight is implemented.

### DIFF
--- a/Uchu.StandardScripts/AvantGardens/SkipBlockYardMissions.cs
+++ b/Uchu.StandardScripts/AvantGardens/SkipBlockYardMissions.cs
@@ -12,8 +12,8 @@ namespace Uchu.StandardScripts.AvantGardens
     {
         public override Task LoadAsync()
         {
-            var man = Zone.GameObjects.First(g => g.Lot == Lot.CrashHelmutNpc);
-            Listen(man.GetComponent<MissionGiverComponent>().OnMissionOk, async tuple =>
+            var crashHelmut = Zone.GameObjects.First(g => g.Lot == Lot.CrashHelmutNpc);
+            Listen(crashHelmut.GetComponent<MissionGiverComponent>().OnMissionOk, async tuple =>
             {
                 var (missionId, _, _, player) = tuple;
 
--- a/Uchu.StandardScripts/AvantGardens/SkipBlockYardMissions.cs
+++ b/Uchu.StandardScripts/AvantGardens/SkipBlockYardMissions.cs
@@ -1,0 +1,49 @@
+using System.Linq;
+using System.Threading.Tasks;
+using Uchu.Core.Resources;
+using Uchu.World;
+using Uchu.World.Scripting.Native;
+using Uchu.World.Social;
+
+namespace Uchu.StandardScripts.AvantGardens
+{
+    [ZoneSpecific(1100)]
+    public class SkipBlockYardMissions : NativeScript
+    {
+        public override Task LoadAsync()
+        {
+            var man = Zone.GameObjects.First(g => g.Lot == Lot.CrashHelmutNpc);
+            Listen(man.GetComponent<MissionGiverComponent>().OnMissionOk, async tuple =>
+            {
+                var (missionId, _, _, player) = tuple;
+
+                // 'go to block yard' mission
+                if (missionId != (int) MissionId.BackupBulwark)
+                    return;
+
+                // small delay to make sure the mission has been added to the player's mission inventory and started
+                await Task.Delay(1000);
+
+                // complete 'go to block yard' mission
+                var missionInventory = player.GetComponent<MissionInventoryComponent>();
+                var bulwarkMission = missionInventory.GetMission(MissionId.BackupBulwark);
+                await bulwarkMission.CompleteAsync();
+
+                // complete 'smash spider queen' mission
+                var arachnophobiaMission = await missionInventory.AddMissionAsync((int) MissionId.Arachnophobia, player);
+                await arachnophobiaMission.StartAsync();
+                await arachnophobiaMission.CompleteAsync();
+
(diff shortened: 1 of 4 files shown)


Uchu.StandardScripts/AvantGardens/SkipBlockYardMissions.cs
```diff
@@ -0,0 +1,49 @@
...
+                var bulwarkMission = missionInventory.GetMission(MissionId.BackupBulwark);
+                await bulwarkMission.CompleteAsync();
+
+                // complete 'smash spider queen' mission
+                var arachnophobiaMission = await missionInventory.AddMissionAsync((int) MissionId.Arachnophobia, player);
+                await arachnophobiaMission.StartAsync();
+                await arachnophobiaMission.CompleteAsync();
+
+                // start 'talk to sky lane' mission
+                var skyLaneMission = await missionInventory.AddMissionAsync((int) MissionId.CheckInwithSkyLane, player);
```
TheNexusAvenger, line 38: Interesting that the script only does the minimum missions to continue instead of all the Block Yard missions (like placing models). Have you made sure no missions have them as prerequisites?

Uchu.StandardScripts/AvantGardens/SkipBlockYardMissions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Uchu.StandardScripts.AvantGardens
+{
+    [ZoneSpecific(1100)]
+    public class SkipBlockYardMissions : NativeScript
+    {
+        public override Task LoadAsync()
+        {
+            var man = Zone.GameObjects.First(g => g.Lot == Lot.CrashHelmutNpc);
+            Listen(man.GetComponent<MissionGiverComponent>().OnMissionOk, async tuple =>
```
TheNexusAvenger: Why is the `async tuple` broken up below instead of `async (missionId, _, _, player)`? I have never seen that before.

Uchu.StandardScripts/BlockYard/BlockYardWarning.cs
```diff
@@ -0,0 +1,20 @@
...
+using Uchu.World.Social;
+
+namespace Uchu.StandardScripts.BlockYard
+{
+    [ZoneSpecific(1150)]
+    public class BlockYardWarning : NativeScript
+    {
+        public override Task LoadAsync()
+        {
+            Listen(Zone.OnPlayerLoad, async player =>
```
TheNexusAvenger, line 12: I like that this message is displayed to the user rather than just blindly complete the missions for the user. `/testmap` requires Mythran permissions though, which not everyone will grant themselves.

Uchu.StandardScripts/BlockYard/BlockYardWarning.cs
```diff
@@ -0,0 +1,20 @@
...
+            {
+                await UiHelper.AnnouncementAsync(player, "Not yet implemented",
+                    "Block Yard is currently not implemented. You can return to Avant Gardens by sending <font color=\"#FF7F00\">/testmap 1100</font> in the chat.");
+            });
+            throw new System.NotImplementedException();
```
TheNexusAvenger: Any reason `NotImplementedException` is being thrown here instead of... not?

TheNexusAvenger (commented):
I'll be testing the functionality later today. Got a few concerns first.

Uchu.StandardScripts/BlockYard/BlockYardWarning.cs
TheNexusAvenger: Tested it and it actually doesn't matter that much. Didn't realize the skipping missions happens in Avant Gardens, not Block Yard.

enteryournamehere: There's no other way to get out of Block Yard besides that command, so unless we fix that launchpad or disable the one in AG I don't think this can be improved

TheNexusAvenger: I guess all we can hope is that players don't go to Block Yard. The reason to go no longer really exists because of skipping the missions in Avant Gardens.

Uchu.StandardScripts/AvantGardens/SkipBlockYardMissions.cs
enteryournamehere: Afraid that just does not work. The delegate must take one parameter, and that would be four. (`OnMissionOk` is an `Event<(int missionId, bool isComplete, MissionState state, GameObject responder)>`). Valid would be `async ((int missionId, bool complete, MissionState state, GameObject player) params)` and then referring to the values using e.g. `params.missionId`, but I don't really see the advantage of that.

Uchu.StandardScripts/BlockYard/BlockYardWarning.cs
enteryournamehere: Er, no. Not sure how that ended up there. Thanks!

Uchu.StandardScripts/AvantGardens/SkipBlockYardMissions.cs
enteryournamehere: Yes, of course. The missions in between are only prerequisites for the directly following mission in the chain. The Arachnophobia mission's completion state determines whether the player can interact with the various NPCs in the Avant Gardens picnic area & whether the Spider Queen effects (dark clouds, Spider Queen in the distance, music) appear in that area.

TheNexusAvenger (approved):
Great temporary solution until we get properties solved. Now a player can go start -> picking a faction and beyond without `/complete`!

PR #300, opened 2021-09-17
Likely fix bug where missions table fails to load

enteryournamehere:
Currently a bug exists where sometimes the Missions table in the cdclient fails to load. This results in `ClientCache.Missions` and `ClientCache.Achievements` being empty, which in turn results in achievements not being completable, and possibly other issues.

This change fixes that bug - it's not entirely clear _why_, or why the issue occurs in the old code, but it seems to work. The old code with a small delay added before `ClientCache.cs:88` (not sure if that's relevant) resulted in the issue occuring 14 out of 25 times, and this new code had it 0 out of 25 times.


TheNexusAvenger (approved):
This issue has shown up before in other ways, but no clue what causes it. Overall, moving the timers to be inside the lock (not sure why they weren't) should hopefully prevent it from being due to any sort of race condition.

PR #304, opened 2021-10-08
Added a tip to start Uchu as an administrator if port binding fails

guyguy2001:
When helping someone over at discord (I'm MasterStormer), they one of the messages I edited, and the solution to the problem was to run Uchu as an admin - so I believe it would be good to add it here


TheNexusAvenger:
We talked about this internally a bit and never responded to this here (oops). There are concerns about broadly recommending running as admin in general for this, but especially because it isn't clear why this happened. Not much exists in Uchu Community about why this happened, and the search results I get about this are only showing for printers.

TheNexusAvenger:
Without any further discussion, I don't feel comfortable blindly telling users to run the server as admin. Nothing stops this pull request from being re-opened with more discussion, but I don't think we should continue with it.

PR #307, opened 2021-10-28
Build file for vscode

teccheck:
I'm using vscode and I've created a `tasks.json` for quickly building Uchu. Does anybody want/need that?


enteryournamehere:
I don't need it personally, but if it's useful to those using VS Code, it seems like a good thing to have

Jettford:
Isn't this auto generated when you select the language?

PR #309, opened 2021-11-01
Racing

### LINKED ISSUE #316: Minigame queue doesn't sync
Only sometimes will other players display in the queue (or flash in and out), but you can still play if you all press play.

teccheck:
Left to do
* [x] Boost behavior
* [x] Ranking updates throughout race (existing checkpoint detection system can be used for this)
* [x] End of the race:
  * [x] Display the correct times & ranking in leaderboard
  * [x] Show ranking after race
  * [x] Add handlers for the buttons there
  * [x] Give rewards
  * [x] Close the zone instance
* [x] Return to correct zone + location for each racing world (currently always goes to NS)
* [x] Add check to kill cars that have fallen off map (which can happen for example in the FV one)
* [x] Investigate into random kill bug
* [x] Correctly handle players trying to join after race has started
* [x] Progress missions and achievements
  * [x] Fix Velocity Lane (you can claim rewards without finishing missions)
  * [x] Find out why some achievements don't work
  * [x] Call `RacingSmashAsync`
  * [x] Call `RacingMissionCompleteAsync`
* [x] Detect player going the wrong way and reset their position
* [x] Respawning
  * [x] Make it work
  * [x] Add small delay (check old footage for this)
* [x] Imagination orb sync
* [x] Configure Server to not accept any more players after race started
* [x] Wait for players in the minigame queue (optional?)
  * [x] Communicate players from queue to RacingControlComponent
* [ ] (Optional) Gameplay config for minPlayerCount
* [ ] What are these [lap colliders](https://explorer.lu-dev.net/objects/9534/5)?
  * [ ] Implement them? 
* [x] Handle players leaving

Scripts:
* [x] All: Imagination crates
* [x] FV: Imagination Dragon Eggs
  * [x] Give imagination
  * [x] Some sort of respawning management script
* [x] FV: Geyser things in the refinery
* [x] FV: Moving platforms (changing paths each lap)

Misc:
* [x] Description for `_arrayNoLenght`
* [ ] Revert testing values
  * [x] LapCount (3)
  * [ ] MinPlayerCount in RacingTask (3)


Jettford:
I ask that you please credit Simon in this code (apologies if I missed you already doing so), he has already has his racing code reimplemented without credit

Jettford:
and I missed it..... thanks for leaving him credit, it was a lot of work that he put into racing.

Uchu.Core/Extensions/RakConnectionExtensions.cs
```diff
@@ -16,7 +16,7 @@ public static void Send(this IRakConnection @this, ISerializable serializable)
             if (serializable == null)
                 throw new ArgumentNullException(nameof(serializable), 
                     ResourceStrings.RakConnectionExtensions_Send_StreamNullException);
-            Logger.Debug($"Sending {serializable}");
+            // Logger.Debug($"Sending {serializable}");
```
TheNexusAvenger, line 19: Is there a reason why several debug statements were taken out? This applies to several files but I am only putting 1 comment for it.

Uchu.Core/Serializable/Struct/Property/PacketProperty.cs
```diff
@@ -88,6 +92,8 @@ public class PacketProperty : IPacketProperty
         /// </summary>
         private Type _arrayLengthPropertyType = typeof(uint);
 
+        private bool _arrayNoLenght = false;
```
TheNexusAvenger, line 99: Can a docstring be added to this to be consistent with the surrounding attributes?

Uchu.World/Objects/Components/ReplicaComponents/RacingControlComponent.cs
```diff
@@ -1,21 +1,713 @@
...
+                BestLapTime = (float) info.BestLapTime.TotalSeconds,
+                RaceTime = (float) info.RaceTime.Elapsed.TotalSeconds,
+            }).ToArray();
+
+            // TODO
+
+            return packet;
+        }
+
+        private enum RacingStatus {
```
TheNexusAvenger: I assume these classes will be moved out like the other packets at some point?

Uchu.World/Objects/Components/ReplicaComponents/VehiclePhysicsComponent.cs
```diff
@@ -1,24 +1,121 @@
...
+                IsTeleporting = true,
+            };
+            packet.VehicleFrameStatsTeleportInfo = teleportInfo;
+            return packet;
         }
     }
-}
\ No newline at end of file
+
+    public struct VehiclePhysicsConstruction {
```
TheNexusAvenger, line 55: I assume these classes will be moved out like the other packets at some point?

Uchu.World/Objects/Components/ReplicaComponents/VehiclePhysicsComponent.cs
```diff
@@ -1,24 +1,121 @@
...
+        // write as byte !
+        public EndOfRaceBehaviorType EndOfRaceBehaviorType { get; set; }
+        public bool IsInputLocked { get; set; }
+        public bool WheelLockExtraFriction { get; set; }
+    }
+
+    public struct VehiclePhysicsSerialization {
+        [Default]
+        public VehicleFrameStatsTeleportInfo VehicleFrameStatsTeleportInfo { get; set; }
+        // [Default] bool with flag? should this be here
```
TheNexusAvenger, line 67: `bool`s having the `[Default]` attribute is supported, because "LU moment".

Uchu.World/Packets/Replica/Serialize/ModuleAssemblySerialization.cs
```diff
@@ -0,0 +1,26 @@
+using Uchu.Core;
+
+namespace Uchu.World
+{
+    public struct ModuleAssemblyConstruction
+    {
+        public bool Flag { get; set; }
+        [Requires("Flag")]
+        public ModuleAssemblyInfo ModuleAssemblyInfo { get; set; }
```
TheNexusAvenger, line 9: Any reason the flag is specified and required instead of just:
```C#
[Default]
public ModuleAssemblyInfo ModuleAssemblyInfo { get; set; }
```

TheNexusAvenger (commented):
Really like where this is going. While this isn't done, I am posting a partial review of the current code now.

Uchu.Core/Extensions/RakConnectionExtensions.cs
enteryournamehere: For racing it's constantly serializing people's positions which makes the logs kind of unreadable

TheNexusAvenger: I see. In that case, I see 2 options: we remove the logging statements altogether, or we start using the `Trace` log level.

Uchu.World/Packets/Replica/Serialize/ModuleAssemblySerialization.cs
enteryournamehere: That seems to be broken for struct properties as are used here, see https://discord.com/channels/640928584113979402/640928584113979404/909078632268443660

Uchu.Core/Serializable/Struct/Property/PacketProperty.cs
teccheck: Sure. I'll add that to the TODO list

Uchu.World/Objects/Components/ReplicaComponents/VehiclePhysicsComponent.cs
teccheck: I think that would be good

Uchu.World/Objects/Components/ReplicaComponents/RacingControlComponent.cs
teccheck: I'm not sure about this one. This is supposed to be an internal thing. It has no use outside of RacingControlComponent iirc

teccheck:
Can we merge this? I'd really like to finish this

enteryournamehere:
IIRC, there were some lingering issues with, uh, communicating position updates with multiple players, or loading of objects, or something, I think? It's been a good while. Either way, though, pretty much all of racing is implemented, so it's fine by me to merge as is :+1:

teccheck:
I'd say we can improve racing later

PR #311, opened 2021-11-11
Upgrade .NET and EF Core to 6.0, C# to 10

enteryournamehere:
As the title says!
Additionally: you can now run server instances as threads instead of separate processes, making it possible to take advantage of .NET 6's Hot Reload functionality to make code modifications while the program is running.
To do this, build in Debug mode (in Release mode, Uchu.Master won't reference Uchu.Instance) and set `DllSource`->`StartInstancesAsThreads` to `true` in the config. (Maybe this isn't the ideal location for this config variable. I put it here so that it's right below `Instance`, but feel free to suggest another place.)
Support for running instances as threads isn't perfect (I think the "auto-shutdown inactive worlds" functionality won't work, and console output won't have the indicator of the source (`[Auth]`, `[Char]`, etc)), but it should be good enough for development purposes.

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Setup .NET
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 5.0.x
+        dotnet-version: 6.0.x
     - name: Restore dependencies
       run: dotnet restore
     - name: Build
--- a/Configuration.md
+++ b/Configuration.md
@@ -42,3 +42,7 @@ A Single Sign-On server makes it possible for users to have an account in one si
 ## Server behaviour
 _`<ServerBehaviour>`_  
 If `<PressKeyToExit>` is set to `true`, when the server process exits it will wait for you to press any key. This prevents windows from disappearing too quickly to read an error message.
+
+## Debug options
+_`<Debugging>`_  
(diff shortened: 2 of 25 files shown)


TheNexusAvenger:
Interesting set of changes to work with .NET 6.0's "Hot Reload" addition. I hope it proves to be a huge timesaver. Beyond that, is there a reason to explicitly specify the C# version we are using the project files if C# 10 is the default now?

> Maybe this isn't the ideal location for this config variable. I put it here so that it's right below Instance, but feel free to suggest another place.

I didn't give an approval because I am not a huge fan of the current location. I can't think of another space that makes sense. Maybe create a new section for debug-related items?

Frozenreflex (approved):
I don't personally have anything wrong with the PR, but if we're gonna add more debug options to the config in the future, what Nexus says seems valid. I'll give my approval anyway.

enteryournamehere:
> is there a reason to explicitly specify the C# version we are using the project files if C# 10 is the default now?

Nope, I just find-and-replaced version 8 to version 10, without realizing it's not necessary to explicitly specify it at all. I've now removed the line from the project files that had it.

> I didn't give an approval because I am not a huge fan of the current location. I can't think of another space that makes sense. Maybe create a new section for debug-related items?

That sounds like a good idea. I've moved the option to a new `<Debugging>` category.

TheNexusAvenger (approved):
Changes seem fine to me. We do need to update UchuDocker as well though.

PR #320, opened 2022-01-06
New Verify API | Changed "account/list"

ArkShocer:
These changes add the functionality to create a web-app for users / administrators to login and manage the servers.
	
Tasks:
- [x] Add API to verify password
- [x] Change output of `account/list` call `usernames` instead of `ids`
- [ ] Add API Key in `Header` to instances
	- [ ] Instances can't connect to master server due to missing api key in header
- [x] Deny access to api calls without key
- [x] Exception for /verify
	- [x] Respond with 401 if no key
	- [x] Verify Response
		- [ ] Respond with key from `config.xml`
- [x] Add New XmlElement `Key` to UchuConfiguration
	- [x] Add `Key` from `config.xml` to `ApiManager.cs`
	- [ ] Add `Key` from `config.xml` to `AccountCommands.cs`
- [x] Set `Access-Control-Allow-Origin, *` for all responses
- [ ] Fix https protocol not working
	- [ ] All instances stop working if changing `http` to `https` in `config.xml`


TheNexusAvenger:
What is the justification/use case for these changes?

teccheck:
I think the verify API could be handy for logging into a web panel like the one Darkflame has.

ArkShocer:
As @TecCheck  mentioned, this API change adds the ability to verify user credentials over http which can then be used to create a web based interface for users/administrators to login to and see their stats/manage the server.

PR #321, opened 2022-01-13
Fix Achievement Progression Issues with Multiple Task Types

### LINKED ISSUE #260: Maelstrom Trench pirates don't smash well
When fighting Stromling Pirates in GF's Maelstrom Trench for the first play session, most will not drop loot and respawn, although they count towards achievements and missions (not every time).

TheNexusAvenger:
Fixes #260

This pull request resolves an issue with `StartUnlockableAchievementsAsync` where null objects would be passed. This is due to a behavior with casing using `as` making the value `null` if the type is incorrect. This affects missions like "Pirate Legend" which involves a Smash task, Script task (...because of course...), and Obtain Item task. Combinations like this would result in the `progress` callback being called with `null` values, leading to an exception when trying to progress a `null` mission task.

### DIFF
--- a/Uchu.World/Objects/Components/Player/MissionInventoryComponent.cs
+++ b/Uchu.World/Objects/Components/Player/MissionInventoryComponent.cs
@@ -787,7 +787,8 @@ namespace Uchu.World
 
                 foreach (var task in mission.Tasks)
                 {
-                    await progress(task as T);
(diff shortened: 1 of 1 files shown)


enteryournamehere (approved):
Good find! Fix looks good to me.

PR #322, opened 2022-01-23
Fix Configuration Paths on Linux with .NET

### LINKED ISSUE #308: Error when loading standard scripts in release
When I join my server on my raspberrypi I get this error when loading venture explorer. I built the latest Uchu with the `linux-arm64` framework using a modded `publish.py`.

```
[Info ] [Z1000] 192.168.178.25:42720's validating client for world!                                                                        
[Info ] [Z1000] Parsing: maps/01_live_maps/space_ship/nd_space_ship.luz                                                                    
[Info ] [Z1000] Parsing: nd_space_ship.lvl                                                                                                 
[Info ] [Z1000] Parsed: 1000                                                                                                               
[Info ] [Z1000] Starting 1000                                                                                                              
[Info ] [Z1000] Checksum: 0x20B8087C                                                                                                       
[Info ] [Z1000] Collecting objects for 1000                                                                                                
[Info ] [Z1000] Loading native scripts...                                                                                                  
[Error] [Z1000] System.IO.DirectoryNotFoundException: Could not find a part of the path                                                    
                '/home/ubuntu/uchu/lib/lib/Uchu.StandardScripts.dll'.
                at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter)
                at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode)
                at System.IO.FileStream.OpenHandle(FileMode mode, FileShare share, FileOptions options)
                at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize,
                FileOptions options)
                at System.IO.File.ReadAllBytes(String path)
                at Uchu.World.Scripting.Native.NativeScriptPack.ReadAssembly()
                at Uchu.World.Scripting.ScriptManager.LoadNativeScripts()
[Info ] [Z1000] Loading managed scripts...                                                                                                 
[Info ] [Z1000] Loading 145 objects for 1000 
```

It tries to load `Uchu.StandardScripts.dll` from `/home/ubuntu/uchu/lib/lib/`, but the file is in `/home/ubuntu/uchu/lib/`

TheNexusAvenger:
Closes #308

This pull request resolves the Linux releases not being able to load scripts due to the relative path used when .NET is installed. The solution used sets the paths in the configuration to be relative to the configuration file on load.

A test release was used on a bare Linux install with .NET not installed and installed. Windows was also tested with a test release, but only with .NET installed.

### DIFF
--- a/Uchu.Core/Config/ConfigurationPathAttribute.cs
+++ b/Uchu.Core/Config/ConfigurationPathAttribute.cs
@@ -1,0 +1,79 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+
+namespace Uchu.Core.Config
+{
+    [AttributeUsage(AttributeTargets.Property)]
+    public class ConfigurationPathAttribute : Attribute
+    {
+        /// <summary>
+        /// Returns an absolute path relative to a configuration path.
+        /// </summary>
+        /// <param name="configurationLocation">File location of the configuration.</param>
+        /// <param name="relativePath">Relative path to make absolute.</param>
+        /// <returns>Absolute path of the relative path.</returns>
+        public static string ReplaceRelativePath(string configurationLocation, string relativePath)
+        {
+            // Normalize the relative path.
+            configurationLocation ??= "";
+            relativePath ??= "";
+            relativePath = relativePath.Replace(configurationLocation.Contains('/', StringComparison.CurrentCulture) ? "\\" : "/", configurationLocation.Contains('/', StringComparison.CurrentCulture) ? "/" : "\\", StringComparison.CurrentCulture);
+            
+            // Return the replaced path.
+            if (Path.IsPathRooted(relativePath)) return relativePath;
+            return Path.Join(Path.GetDirectoryName(configurationLocation), relativePath);
+        }
+        
+        /// <summary>
+        /// Replaces local configuration paths relative to the configuration file.
+        /// </summary>
+        /// <param name="configurationLocation">File location of the configuration.</param>
+        /// <param name="configurationObject">Object to replace the paths in.</param>
+        public static void ReplaceFilePaths(string configurationLocation, object configurationObject)
+        {
+            // Return if there is no object to replace.
+            if (configurationObject == null)
+            {
+                return;
+            }
+        
+            // Iterate over the properties.
+            foreach (var property in configurationObject.GetType().GetProperties())
+            {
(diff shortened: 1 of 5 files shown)


Uchu.Core/Config/ConfigurationPathAttribute.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// </summary>
+        /// <param name="configurationLocation">File location of the configuration.</param>
+        /// <param name="relativePath">Relative path to make absolute.</param>
+        /// <returns>Absolute path of the relative path.</returns>
+        public static string ReplaceRelativePath(string configurationLocation, string relativePath)
+        {
+            // Normalize the relative path.
+            configurationLocation ??= "";
+            relativePath ??= "";
+            relativePath = relativePath.Replace(configurationLocation.Contains('/', StringComparison.CurrentCulture) ? "\\" : "/", configurationLocation.Contains('/', StringComparison.CurrentCulture) ? "/" : "\\", StringComparison.CurrentCulture);
```
enteryournamehere, line 22: Is this necessary, can't Windows also handle forward slashes?

teccheck: I think no, but I know that java provides a char of the os's path separators. Maybe c# has that as well?

TheNexusAvenger: I am not sure if this is a requirement. In the past, I have gotten some weird behavior *not* normalizing paths, so this is more preventative. To confirm this, I would need to test without this on both Windows and Linux.

enteryournamehere: On Linux you'll need to replace `\\` with `/` but I think Windows can handle either. But if it works it works, no need to test everything again for a tiny change imo

TheNexusAvenger: I *think* the initial reason I did this is because of `Path.Combine`. Having a mix of `\` and `/` may have caused issues for me in the past with *something*, but I don't remember what.

enteryournamehere (approved):

teccheck:
I tested the changes on my Raspberry Pi and everything seems to work :+1:

PR #323, opened 2022-01-30
Fix Interrupting Quickbuilds

### LINKED ISSUE #269: quickbuilds are hard to finish if shift is let go
If you let go of shift early during a quickbuild, in many cases the shift prompt disappears (sometimes reappearing in a different place), making it challenging to get it to finish before it respawns. The player might get stuck on an animation afterwards, and it makes all shift interactions unusable unless you smash yourself.

TheNexusAvenger:
Closes #269

This pull request adds missing messages for interrupting quickbuilds. If a player were to let go of the interact button or run out imagination, they would be unable to interact with anything in the game. These extra messages fix that, and are the ones used in DLU.


enteryournamehere (approved):
Great :+1:

PR #324, opened 2022-01-31
Add Crux Prime Enemy Spawning

### LINKED ISSUE #229: There are no enemies in Crux Prime besides bosses
I know they were taken out at one point in order to reduce lag, but I think the server has evolved to be stable enough to bring them back without lag being an issue.

TheNexusAvenger:
Closes #229

This pull request implements enemy spawning on Crux Prime. The implementation used is ported over from Darkflame LEGO Universe with some tweaks due to how Uchu works.

Named enemies are included in this implementation.


teccheck (approved):
I can't test these changes but the code looks good to me :+1:

PR #326, opened 2022-02-09
Persist Selected Consumables

### LINKED ISSUE #192: Selected Skills In Action Bar Don't Persist
image
When placing items in the 5th slot of the Action Bar, they can be used, but the selection doesn't persist when re-logging in. The [`SetConsumeableHandler`](https://github.com/UchuServer/Uchu/blob/c16eba45265dc84e37cec189182998540b50e7f2/Uchu.World/Handlers/GameMessages/SkillHandler.cs#L41) is implemented but appears to set a value that is never read or saved.

TheNexusAvenger:
Closes #192

This pull request:
- Makes selected consumables persist between sessions. It just wasn't fully implemented.
- Adds a helper script for creating database migrates in the future, which was used for this pull request.
  - Added a doc for academic requirements as well as covering the script above.

### DIFF
--- a/Uchu.Core/Database/Models/Character.cs
+++ b/Uchu.Core/Database/Models/Character.cs
@@ -184,6 +184,8 @@ namespace Uchu.Core
         public bool LandingByRocket { get; set; }
         
         public int InventorySize { get; set; }
+        
+        public int SelectedConsumable { get; set; }
 
         public int VaultInventorySize { get; set; }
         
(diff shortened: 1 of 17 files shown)


teccheck (approved):
I tested the changes and they worked. The helper scripts are nice and I think they'll help a lot

Frozenreflex (approved):

Issue #291, opened 2021-08-20
Vendor Inventories Persist Between Sessions


TheNexusAvenger:
Inventories are currently saved in the database with no checks to the inventory type ([current save code](https://github.com/UchuServer/Uchu/blob/35d426c0ab80f8767e7e239f1c742ada19cc5fd7/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs#L123)). This results in [`InventoryType`](https://github.com/UchuServer/Uchu/blob/c064bd2240a5a16c3b0ff5a8c5803a826d72c968/Uchu.Core/Enums/InventoryType.cs)s that shouldn't persist, like `VendorBuyback` (11) saving. This results in the vendor buyback slots persisting positions but not displaying items, and a lot of extra items being stored in the database for every player.

PR #329, opened 2022-04-04
Filter Vendor Buyback Items

zxc8027:
For the 2nd of my 3 contributions for a class, I took a look at #291. In order to resolve it, I figured that:
1. Vendor items should not be loaded from the database.
2. Vendor items should not be saved to the database.
3. Vendor items should not be counted towards missions.

The `Items` array isn't referenced that much in the code. My hope is that I didn't miss anything.

### DIFF
--- a/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
+++ b/Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
@@ -129,7 +129,7 @@ namespace Uchu.World
         public async Task SaveAsync(UchuContext context)
         {
             if (!this._loaded) return;
-            var itemsToSave = Items;
+            var itemsToSave = UnsoldItems;
 
             var character = await context.Characters.Where(c => c.Id == GameObject.Id)
                 .Include(c => c.Items)
@@ -148,7 +148,6 @@ namespace Uchu.World
             {
                 var savedItem = character.Items.FirstOrDefault(i => i.Id == itemToSave.Id);
                 var inventoryType = itemToSave.Inventory?.InventoryType ?? InventoryType.None;
-                if (inventoryType == InventoryType.VendorBuyback) continue;
                 if (savedItem == default)
                 {
                     character.Items.Add(new InventoryItem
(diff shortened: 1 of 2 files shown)


Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
```diff
@@ -140,7 +146,9 @@ public async Task SaveAsync(UchuContext context)
             // Update existing items and add new items
             foreach (var itemToSave in itemsToSave)
             {
-                var savedItem = character.Items.FirstOrDefault(i => i.Id == itemToSave.Id); 
+                var savedItem = character.Items.FirstOrDefault(i => i.Id == itemToSave.Id);
+                var inventoryType = itemToSave.Inventory?.InventoryType ?? InventoryType.None;
+                if (inventoryType == InventoryType.VendorBuyback) continue;
```
TheNexusAvenger: `itemToSave` directly points to `Items` with no modifications. Since you are filtering sold items out, it looks like you can just use the `UnsoldItems` property you added.

TheNexusAvenger (commented):
The changes seem mostly fine for me expect 1 comment. This confirmed my theory that not loading the vendor items fixes items being sold appearing random spots.
image

A feature missing from this is updating the collect mission tasks when selling items. I have no clue how big this would be since mission tasks reverting isn't a feature, and I fear this would break some achievements that rely on collecting total items. Someone who knows more about this can correct me if this wasn't a feature in the live game. If your school assignment allows for considering this out of scope, we can merge it after a response for the comment and open a GitHub Issue to track this new part.

Uchu.World/Objects/Components/Player/InventoryManagerComponent.cs
zxc8027: That change makes sense to me. I will need to reach out to my professor to see if I can consider this done or not.

zxc8027:
> A feature missing from this is updating the collect mission tasks when selling items. I have no clue how big this would be since mission tasks reverting isn't a feature, and I fear this would break some achievements that rely on collecting total items. Someone who knows more about this can correct me if this wasn't a feature in the live game. If your school assignment allows for considering this out of scope, we can merge it after a response for the comment and open a GitHub Issue to track this new part.

The professor confirmed I can get credit for this being merged as-is. If the issue isn't clear, I don't think I should look into it.

TheNexusAvenger (approved):

enteryournamehere (approved):
Looks good to me, thanks for the PR!

PR #331, opened 2022-04-06
added command for toggled speed boost

AbeLab404:
Simple command addition for quality of life change
/speed will constantly refresh a speed boost until toggled off

### DIFF
--- a/Uchu.World/Handlers/Commands/CharacterCommandHandler.cs
+++ b/Uchu.World/Handlers/Commands/CharacterCommandHandler.cs
@@ -17,6 +17,8 @@ using Uchu.World.Client;
 using Uchu.World.Filters;
 using Uchu.World.Scripting.Native;
 using Uchu.World.Social;
+using Uchu.World.Objects;
+using System.Threading;
 
 namespace Uchu.World.Handlers.Commands
 {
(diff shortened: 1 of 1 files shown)


TheNexusAvenger (changes requested):
The idea is interesting, but I am not a fan of the approach.
1. `toggleSpeed` is global. While the command only applies to the calling player, the toggle is global and can be affected by another player.
2. The approach for applying the speed boost uses a loop with skills. That doesn't seem like a proper solution.

For testing, we typically use `/fly on (speed)` to quickly get around the map. This also allows us to go vertically instead of just on the normal play area.

AbeLab404:
Understood @TheNexusAvenger, I was not aware of /fly on (speed). That definitely seems like a more efficient method.

I'm still pretty new to Uchu but I'll keep looking into things to work on.

NinjaOfLU:
If I might add something: I _do_ think that there's some room for a command to set a player's speed, separate to the jetpack. I don't know how applicable it would be for those adding new features, but for people testing worlds and the like (and I think Uchu would be my go to for initially testing modded worlds), I could see a command to change speeds being helpful (there are just certain things you can do whilst walking which you can't, whilst flying).

I don't know how the speedboosts in jetpacks are coded, and am not in a position to give comments on the code. As for how one _might_ apply a permanent speedboost, I feel like there are a few avenues into which you could delve. For instance, iirc from skill trees, the buff system used in item skill trees currently provides the ability to activate/deactivate buffs on the clientside. I don't know how easy it is to send an arbitrary skill-tree from the serverside, but that might be a way of implementing things to look into. There's also some WorldConfig stuff in cdclient which changes movement speeds. Again, I don't know if that's something which can be sent from the server - mostly just trying to think of the the systems in LU which most naturally do what you want. I know certain times the server can somewhat override what's in the client database, but have no idea if that's relevant here.

PR #332, opened 2022-04-10
Fix /complete Command Breaking Mission Progression

zxc8027:
While working towards a different issue, I used the `/complete` command to progress a mission and became unable to get new missions. The problem is missions get into the completed state but the tasks don't. Changing the `Completed` property to check for the state resolved the issue for me.


TheNexusAvenger (approved):
Can confirm this is an issue and this resolves it. Bit annoyed that I missed that before because I am pretty sure this appeared when I added random mission pools.

PR #333, opened 2022-04-23
Ignore Ghosting Filter for Gnarled Forest Special Case

### LINKED ISSUE #264: Brig Rock loads in super late

TheNexusAvenger:
Closes #264

This pull request attempts to resolve Brig Rock loading late with a [fairly simple workaround used by DLU](https://github.com/DarkflameUniverse/DarkflameServer/blob/80a09881c097a07520914b4f73ea1aa7970baba7/dGame/Entity.cpp#L738).  Ideally, a more complex check should be done based on the distance to the boundary of the object. However, this is a lot simpler and is *probably* fine with more recent hardware.

### DIFF
--- a/Uchu.World/Objects/GameObjects/GameObject.cs
+++ b/Uchu.World/Objects/GameObjects/GameObject.cs
@@ -120,6 +120,8 @@ namespace Uchu.World
 
         public ReplicaComponent[] ReplicaComponents => Components.OfType<ReplicaComponent>().ToArray();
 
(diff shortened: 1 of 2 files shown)


teccheck:
DLU checks for 2 components while your implementation checks for 3. Is this intentional?

TheNexusAvenger:
Yes. We handle Trigger components differently compared to DLU and Brig Rock has this.

enteryournamehere (approved):
The check does look a bit odd but it seems like a decent workaround, it could always be improved later on.

PR #334, opened 2022-04-23
Add Release Document

TheNexusAvenger:
To wrap up my course work, my professor wanted me to write a document about creating releases. This pull request covers adding that document as well as adding new releases.

### DIFF
--- a/docs/releases.md
+++ b/docs/releases.md
@@ -1,0 +1,64 @@
+# Releases
+This document covers making releases for distribution.
+
+## Windows Warning
+The process for creating releases is highly automated and can be run on
+any platform. However, while the releases built on macOS/Linux for Windows work
+as expected, the releases built on Windows for macOS/Linux require changing
+file permissions to run. This is because Linux file permissions don't exist
+on Windows and default to not being executable on macOS and Linux. **Publishing
+macOS and Linux releases built on Windows are strongly discouraged.**
+If you are on Windows, a Linux virtual machine, such as [Windows Subsystem for Linux (WSL)](https://docs.microsoft.com/en-us/windows/wsl/about)
+or [Oracle VirtualBox](https://www.virtualbox.org/).
+
+## Setup
+The only supported way to create releases for automation tools [like UchuTool](https://github.com/UchuServer/UchuTool)
+using the [`publish.py`](../publish.py) script. In order to use it, 2 programs
+need to be installed:
+1. [Microsoft .NET 6](https://dotnet.microsoft.com/en-us/download/dotnet/6.0) for building releases.
+2. [Python 3](https://www.python.org/downloads/) for running the helper script.
+
+See any operating system-specific instructions for setup. No additional setup
+of Python libraries is required to run the script. However, you will need to
+make sure you have `dotnet` in your `PATH` environment variable. You can check
+by opening a command prompt/terminal and entering `dotnet --version`.
+
+## Building Releases
+With Python and .NET set up, the recommended process for creating releases is
+the following:
+1. Open a command prompt or terminal in the root Uchu directory.
+2. Run `python publish.py` or `python3 publish.py` depending on your system.
+
+How long it takes depends on your system. At most, it should take a few minutes.
+After the script completes, there will be a directory named `bin` created with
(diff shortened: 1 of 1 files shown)


docs/releases.md
```diff
@@ -0,0 +1,64 @@
...
+```python
+PLATFORMS = [
+    ["Windows-x64", "win-x64"],
+    ["macOS-x64", "osx-x64"],
+    ["Linux-x64", "linux-x64"],
+    ["Linux-ARM64", "linux-arm64"], # Will create a 64-bit ARM release for Linux (linux-arm64) as Uchu-Linux-ARM64.zip
+]
+```
+
+It is recommended to do the same for [like UchuTool](https://github.com/UchuServer/UchuTool)
```
teccheck: I'd replace "like UchuTool" with "UchuTool". It's probably a copy-paste thing ;)

teccheck:
I think this is quite good :+1:

docs/releases.md
TheNexusAvenger: Good catch. that was a mistake.

docs/releases.md
```diff
@@ -0,0 +1,64 @@
+# Releases
+This document covers making releases for distribution.
+
+## Windows Warning
+The process for creating releases is highly automated and can be run on
+any platform. However, while the releases built on macOS/Linux for Windows work
+as expected, the releases built on Windows for macOS/Linux require changing
+file permissions to run. This is because Linux file permissions don't exist
+on Windows and default to not being executable on macOS and Linux. **Publishing
+macOS and Linux releases built on Windows are strongly discouraged.**
```
enteryournamehere: are strongly discouraged -> is strongly discouraged

docs/releases.md
```diff
@@ -0,0 +1,64 @@
...
+If you are on Windows, a Linux virtual machine, such as [Windows Subsystem for Linux (WSL)](https://docs.microsoft.com/en-us/windows/wsl/about)
```
enteryournamehere, line 11: after the first comma I'd add "you can use", or something similar

docs/releases.md
```diff
@@ -0,0 +1,64 @@
...
+or [Oracle VirtualBox](https://www.virtualbox.org/).
+
+## Setup
+The only supported way to create releases for automation tools [like UchuTool](https://github.com/UchuServer/UchuTool)
+using the [`publish.py`](../publish.py) script. In order to use it, 2 programs
```
enteryournamehere: missing "is" before "using the"

docs/releases.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+It is recommended to do the same for [UchuTool](https://github.com/UchuServer/UchuTool)
+as well. Be aware that the [code for downloading releases](https://github.com/UchuServer/UchuTool/blob/master/Uchu.Tool/Action/Update.cs)
+may need to be updated depending on how it ends up attempting to find a version to download.
+At the time of writing, anything beyond 64-bit x86 (commonly known as x64) is untested,
+and releases specifically for Windows/macOS/Linux versions are untested.
+
+If you want to add new runtimes, submit a pull request. Make sure to test the
+release on the intended system and why this new release is useful.
```
enteryournamehere: I think a "describe" was intended to be between "and" and "why" here

enteryournamehere (changes requested):
There are some sentences where a few words seem to have gone missing, overall it's great though!

docs/releases.md
TheNexusAvenger: I am not sure about this request. I interpret my wording as "these are examples of tools that work, but others exist" while I would interpret your requested change as "these are the tools that work and no others exist or are recommended".

enteryournamehere: It does not have to be exactly "you can use", it's just that "a Linux virtual machine, such as WSL or Virtualbox." isn't a complete sentence (and also doesn't imply to me that there are any other options beside a VM). So my suggestion is to add something to make it one. It could say "it's possible to use a Linux virtual machine", or "one option is a Linux virtual machine", or anything else you like.

TheNexusAvenger: Uh, oops. There should be more words after all that. More has been added.

enteryournamehere: That's perfect thanks :+1:

enteryournamehere (approved):
Nice to have this documented!

PR #335, opened 2022-05-01
Toggle buff to base stats

AbeLab404:
Change to /buffme which allows it to be toggled on/off in order to set previous health, armor, etc.
Previously, there was no way to do this without /smash

This does create a global List, but I could not figure out how else to save the original stats.

### DIFF
--- a/Uchu.World/Handlers/Commands/CharacterCommandHandler.cs
+++ b/Uchu.World/Handlers/Commands/CharacterCommandHandler.cs
@@ -187,20 +187,74 @@ namespace Uchu.World.Handlers.Commands
             return "You smashed yourself";
         }
         
+        List<uint> stats = new List<uint>();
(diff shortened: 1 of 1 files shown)


TheNexusAvenger (changes requested):
This change isn't safe to merge. While it *can* work for singleplayer, the `stats` list is global. If someone were to do `/buffme on` and someone else did `/buffme off`, they will get the stats of the other player. It also seems possible that `/buffme off` can be called without using `/buffme on`, leading to an `IndexOutOfRangeException` on line 233. Additionally, if a player were to buff themselves and change their character, the changes they made to their stats will be reset when unbuffed.

Beyond functionality, I am not a fan of using a list to store the values like this. A `class` or ` struct` is much more suitable for this use case instead of a list of numbers. It is a lot more difficult to interpret and can lead to referencing the incorrect indexes.

PR #347, opened 2023-08-20
Improve user feedback for chat filter

enteryournamehere:
Small fix to improve the feedback users get when their message doesn't get past the chat filter. (See [lu_packets](https://lcdruniverse.org/lu_packets/lu_packets/world/client/struct.ChatModerationString.html), [DLU](https://github.com/DarkflameUniverse/DarkflameServer/blob/main/dNet/WorldPackets.cpp#L199) for reference.)

Before:
image

After:
image


teccheck (approved):
Looks good to me and works. Nothing more to add :+1:

PR #348, opened 2023-08-21
Fix respawning for racing

### LINKED ISSUE #346: Respawn in racing broken

enteryournamehere:
The previous code
```c#
var playerInfo = Players.Find(info => info.Player.Id == player.Id);
playerInfo.Vehicle = car;
```
was broken because `Players` contains `RacingPlayerInfo` structs, which is a value type, not a reference type. This means that `var playerInfo = ...` results in a local copy of the data. The following line then only modifies this local copy, leaving `Players` unchanged.

This new code creates a local copy of the data, modifies it, and then updates the struct stored in the `Players` list.

### DIFF
--- a/Uchu.World/Objects/Components/ReplicaComponents/RacingControlComponent.cs
+++ b/Uchu.World/Objects/Components/ReplicaComponents/RacingControlComponent.cs
@@ -230,9 +230,11 @@ namespace Uchu.World
             RacingPlayerInfo playerInfo;
 
             if (IsPlayerRegistered(player)) {
-                playerInfo = Players.Find(info => info.Player.Id == player.Id);
+                var playerInfoIndex = Players.FindIndex(info => info.Player.Id == player.Id);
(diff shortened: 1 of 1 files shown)


teccheck (changes requested):
In line 233 there's another one of those bugs. To me it seems like it's working anyways but it might be good to fix this as well.

```cs
playerInfo = Players.Find(info => info.Player.Id == player.Id);
playerInfo.Player = player;
playerInfo.PlayerLoaded = true;
```
Other than that it looks good to me. 

I looked into [ref structs](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/ref-struct) but they don't seem to be the right choice

enteryournamehere:
Oh, good catch!

teccheck:
I made this code after all... The more I look at it, the more I want to redo the whole thing xD

teccheck (approved):
Looks good to me. Thank you :+1:

PR #350, opened 2023-08-28
Fix skill used in imagination crate script

enteryournamehere:
Should be skill 586 instead of 585, reference: https://explorer.lu-dev.net/scripts/ai/racing/objects/race_imagine_crate_server.lua

(As a result, the full racing imagination bar is filled instead of only 1/6 points)


Jettford:
May I note just for future, it may be better to link scripts from Arrons repo instead of LU Explorer as LU Explorer removes all comments, it has been known to cause issue.

PR #351, opened 2023-08-31
Modular building rework

teccheck:
This fixes a few issues that my old code had. It's almost a complete rewrite. Not sure if `BuildBorderComponent` is needed.

TODOs:
- [x] Implement pick up mechanic
- [x] Remove log statements
- [x] Remove `BuildMode` enum
- [x] ~~Maybe remove `BuildBorderComponent`~~
- [x] Implement and use `PushEquippedItemState` and `PopEquippedItemState`
- [x] Heal player when popping item state
~~Exit build session if player leaves server~~
- [x] Documentation
  - [x] `ModularBuildingHandler`
  - [x] `ModularBuilderComponent`

### DIFF
--- a/Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
+++ b/Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
@@ -1,250 +1,252 @@
 using System.Linq;
 using System.Threading.Tasks;
-using InfectedRose.Lvl;
 using InfectedRose.Core;
 using Uchu.Core;
-using Uchu.Core.Client;
-using System;
-using System.Collections.Generic;
 
 namespace Uchu.World
 {
     public class ModularBuilderComponent : Component
     {
-        public GameObject BasePlate { get; private set; }
-        
+        public GameObject BuildArea { get; private set; }
+
         public bool IsBuilding { get; private set; }
 
-        public BuildMode Mode { get; private set; }
+        private Lot[] CurrentModel = null;
 
-        private Lot[] CurrentModel { get; set; }
+        private Player Player => (Player)this.GameObject;
 
         /// <summary>
         /// Called when a modular build is completed.
         /// </summary>
         public Event<ModularBuildCompleteEvent> OnBuildFinished { get; }
-        
+
         protected ModularBuilderComponent()
         {
             this.OnBuildFinished = new Event<ModularBuildCompleteEvent>();
         }
 
-        public async Task StartBuildingAsync(StartBuildingWithItemMessage message)
+        public void StartBuild(StartBuildingWithItemMessage message)
         {
-            if (!(GameObject is Player player))
-                return;
-
-            var inventory = GameObject.GetComponent<InventoryManagerComponent>();
-            var thinkingHat = inventory[InventoryType.Items].Items.First(i => i.Lot == 6086);
-
-            await thinkingHat.EquipAsync(true);
-            
+            Logger.Debug($"Associate: {message.Associate}");
             IsBuilding = true;
-            BasePlate = message.Associate;
-            Mode = GetBuildModeForBasePlateLot(BasePlate.Lot); 
+            BuildArea = message.Associate;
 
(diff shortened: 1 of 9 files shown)


EmosewaMC:
as a note, if its not already done, when you pop the equipped items, you need to heal the player to full stats.

teccheck:
I didn't know that. Thanks 👍

teccheck:
I just noticed, that I reformatted `InventoryComponent.cs` and `InventoryHandler.cs`. Is that OK or should I revert that?

teccheck:
I think this is ready for review. If everything works, I'll remove all or most of the log statements

Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
```diff
@@ -1,250 +1,292 @@
...
 
         /// <summary>
-        /// When the player picks up a part from the ground.
+        /// Stops the build mode and cleans up.
         /// </summary>
-        public async Task Pickup(Lot lot)
+        private async Task ExitBuild()
         {
-            if (!(GameObject is Player player)) return;
+            // Should be removed once PopEquippedItemState is implemented
```
enteryournamehere: It seems you have, in fact, implemented PopEquippedItemState, right?

Uchu.World/Objects/Components/ReplicaComponents/InventoryComponent.cs
```diff
@@ -229,13 +233,57 @@ private async Task EnsureItemUnEquipped(Item item)
...
+            if (EquippedItemStack == null || EquippedItemStack.Count == 0)
+                return;
+
+            Logger.Debug($"Equipped Items: {EquippedItems.Length}");
+            Logger.Debug($"Items on stack: {EquippedItemStack.Count}");
+
+            foreach (var item in EquippedItems)
+            {
+                Logger.Debug($"Unequip {item}");
+                await UnEquipItemAsync(item, false);
```
enteryournamehere, line 273: Why does serialize need to be false here? Is it just to avoid sending many messages in rapid succession?

enteryournamehere (approved):
Works great! Just two questions :slightly_smiling_face:

enteryournamehere:
Btw, not a new issue I think, just something to note & be fixed at some point, doesn't need to be now:

If you close your client while building (I did because mine froze, but I couldn't reproduce that) and restart, you can walk around with the thinking cap on & a rocket part in your arms. You can enter + leave the building area again to get rid of it. I think ideally it should call ModularBuilderComponent.ConfirmBuildExit() so that the player gets their original rocket back instead of the loose parts. Screenshot_20230910_163457

Uchu.World/Objects/Components/ReplicaComponents/InventoryComponent.cs
teccheck: Yes that's the biggest reason for it. I also wanted to make sure that the client would not get any messages in the wrong order (not sure if this is necessary).

teccheck:
> I think ideally it should call ModularBuilderComponent.ConfirmBuildExit() so that the player gets their original rocket back instead of the loose parts.

I think, I will implement this as well. It fits the PR.

EDIT: Btw you can just hit `ALT` `F4` to close the game and it will end up in the same state as you described.

Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
teccheck: Oh right. I forgot that. Will remove that and test again

teccheck:
Ok never mind. Currently it's not possible to revert the state of the player once they left because the server crashes when it tries to send messages to that player. For this to work, the server would need to catch null pointer and just not send anyting

Uchu.World/Objects/Components/Player/ModularBuilderComponent.cs
teccheck: Code is now removed

